PR #13, opened 2018-11-02
Allow short aliases to refer to application/* as well as text/* for json types

antony:
This is the rest of my fix - a few extra tests, but also removes the need for lang to always start with `text/`. Often json types use `application/`


kaisermann:
🎉

PR #24, opened 2019-05-05
fix: 🐛 less and stylus stand-alone processor

Cweili:
Fix less and stylus stand-alone processor


kaisermann:
Ooooh, my bad! Thanks 👏  !

Issue #31, opened 2019-05-15
Sapper Typescript Pug Example


burner:
I'am trying to get sapper with rollup to work with typescript and pug, but I just can't wrap my head around the build process.

Is there any change to get a example with the above mentioned tech?

Thank you

laurentpayot:
Regarding Pug, you have to use mixins. Maybe answer can help you: https://stackoverflow.com/a/51809972/2675387

burner:
maybe need to clarify. You don't even get the start sapper project to compile with pug and ts. I don't use and svelte html control structures at this point.

burner:
might also be helpful to have the example rollup.config.js close to the one generated by sapper rollup.config.js

kaisermann:
@burner to preprocess pug and typescript you should just install `typescript`, `pug`, `import preprocess from 'svelte-preprocess'` and pass it to the svelte `preprocess` option:

```
preprocess: preprocess()
```

For typescript, you should also config your `tsconfig.json`.

Once I get some spare time I'll provide a sapper example in the `examples` directory 👍

burner:
@kaisermann thank you for coming back to me. 
I created a PR that shows where I want to go with this.

kaisermann:
@burner Sapper example was added 🎉

burner:
@kaisermann thank you very much. I still got a problem though.
/blog gives me
```
preprocess/examples/sapper_ts_pug_scss/src/routes/blog/index.svelte
'posts' is not defined
```

The page shows a 500 with 
```
ReferenceError: posts is not defined
```

PR #34, opened 2019-05-22
Sappar + ts + pug + scss example

burner:
working on https://github.com/kaisermann/svelte-preprocess/issues/31 to show where I want to go with this.


kaisermann:
@burner I've refactored and added some more code to the sapper example 👍 

Some pug mixins were also added: https://github.com/kaisermann/svelte-preprocess#pug

PR #36, opened 2019-05-25
feat: prepend scss with data property

bastienrobert:
The `data` option in the SCSS preprocessor prevented the contents of the file.
I made a quick edit so now, the `data` attribute in the options is composed of the passed options and the file content.


bastienrobert:
Small precision: it seems that the SCSS files imported with a `@import` in `data` attribute are not hot reloaded.

src/transformers/scss.js
```diff
@@ -7,12 +7,12 @@ module.exports = ({ content, filename, options }) => {
   options = {
     includePaths: getIncludePaths(filename),
     ...options,
+    data: options.data + content,
```
kaisermann: if `options.data` is `undefined` this would prepend a `"undefined"` string to the file content, right?

kaisermann:
Could you clarify why would you need this change? Is it just to prepend some content to your scss files?

src/transformers/scss.js
bastienrobert: yes you're right, my bad 😱

could change this by `options.data ? options.data + content : content`

bastienrobert:
I need this to add some default imports in top of my SCSS files. Usually I import a `config.scss` file with my SCSS vars.

src/transformers/scss.js
kaisermann: That would work, yeah! It would also be great to move this logic to just below the `options = {}`, like on the [`sass-loader`](https://github.com/webpack-contrib/sass-loader/blob/master/lib/normalizeOptions.js#L37) for better readability 👍

bastienrobert:
Sorry for the delay @kaisermann, I just committed to fix it 🙃

kaisermann:
@bastienrobert No problem! Released on `v2.12.0` 🙏 Thanks!

PR #42, opened 2019-06-19
feat: 🎸 global style

hantrongbinh:
### Global style support

Use `globalStyle` preprocess or `global` attribute to apply styles to a selector globally.

```html
<style global lang="scss">
  $color: red;
  div {
    color: $color;
  }
</style>
```

```js
import { scss, globalStyle } from 'svelte-preprocess'

svelte.preprocess(input, [
  globalStyle(),
]).then(...)
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,28 @@ Current supported out-of-the-box preprocessors are `SCSS`, `Stylus`, `Less`, `Co
 </style>
 ```
 
+### Global style support
+
+Use `globalStyle` preprocess or `global` attribute to apply styles to a selector globally.
+
+```html
+<style global lang="scss">
+  $color: red;
+  div {
+    color: $color;
+  }
+</style>
+```
+
+```js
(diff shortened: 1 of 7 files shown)


kaisermann:
First of all, thanks for this! The `global` attribute should be a great feature to add, but I think it should be available for every style tag, not only for `scss`.

hantrongbinh:
@kaisermann 
>The global attribute should be a great feature to add, but I think it should be available for every style tag, not only for scss. 
I think so. But I'm using `scss` in my project so I make for `scss` first

> The global transformation can happen just before handling the final css to svelte.
be more specific please

kaisermann:
As discussed on discord, I think the regex to detect a selector doesn't cover all possible cases. i.e change the test to `<style global>@media(min-width:10px){div{color:red}}.test{}</style>` and you will see that no selector is transformed. Since CSS is not a regular language, I think a better approach would be to use `postcss` and some kind of plugin to transform the selectors.

PR #53, opened 2019-07-27
Fix README index hyperlink

woutdp:
Fixes the hyperlinks in the documentation.


kaisermann:
Hi! Thanks for this PR. Unfortunately, I use a vscode extension to generate the TOC, so the problem is with the extension itself. I've just pushed a commit which moves those links out of the headings :tada:

PR #65, opened 2019-09-10
Make @keyframe at-rules global.

elliotwaite:
Prefix all @keyframe at-rule names with '-global-'.
For example:
  @keyframes slide-in {from{} to{}}
will become:
  @keyframes -global-slide-in {from{} to{}}


kaisermann:
Thank you very much @elliotwaite! Nice fix 😁

PR #84, opened 2019-10-27
Add `sass` installation line to README

igoradamenko:
According to the source code the Sass-transformer works with `node-sass` or `sass`, but there is no notes about it in the README.

I've added installation line just to mark this fact, because it is important. My first impression was that it can't work with `sass`.


kaisermann:
Oh, nice one! Thank you very much 🎉

Issue #98, opened 2019-11-11
Autosubscriptions not working in <script>


cortopy:
**Describe the bug**
As mentioned in https://github.com/kaisermann/svelte-preprocess/issues/81, autosubscriptions work in the html part but not `<script>`

In the following:

```html
<script lang="ts">
	import { count } from './stores.js';
	import Incrementer from './Incrementer.svelte';
	import Decrementer from './Decrementer.svelte';
	import Resetter from './Resetter.svelte';

        let isLarge = $count > 10;
</script>

<h1>Count is now {isLarge ? "large" : "small"}</h1>

<Incrementer/>
<Decrementer/>
<Resetter/>
```
two errors are thrown because (1) count is not used and (2) $count is undefined.

I know this example is pretty dumb because the evaluation could be done in the HTML, but basically using autosubscriptions in the <script> confuses the ts transpiler.

TwitchBronBron:
I realize this is a huge hack, but it might at least get rid of the errors. How about we inject variable declarations into the in-memory copy of the source code right before we preprocess it? Some type of algorithm like this:

```typescript
var autoSubs = [];
var rawTypescriptText = getTextFromScriptTagSomehow();
//regex match every variable that starts with a dollar sign
var regexp = /(\$\w*)/g;
var match;
while (match = regexp.exec(rawTypescriptText)) {
  var name = match[0];
  //add this name if we don't have it yet
  if (autoSubs.indexOf(name) === -1) {
    autoSubs.push(name);
  }
}
//generate the var declaration text. i.e.: "var $item1: any;var $item2: any;var $item3: any;"
var injectedText = autoSubs.reduce(function(cumulativeText, autoSubName){
	return cumulativeText + 'var ' + autoSubName + ': any;';
}, '');

var modifiedTypescriptText = injectedText + rawTypescriptText;
transpile(modifiedTypescriptText);
```

So this code:
```html
<script lang="typescript">
    import { age } from './ageStore;
    var ageNextYear = $age + 1;
</script>
```
would become

```html
<script lang="typescript">
    var $age: any;import { age } from './ageStore;
    var ageNextYear = $age + 1;
</script>
```


If we append to the beginning of the first line, we might need to then fix column offset numbers for any diagnostics that appear on the first line.

This is just the first step, getting rid of the compile errors. Once that's done, perhaps we could do some smarter type detection instead of defaulting to :any. 

[...]

PR #124, opened 2020-03-22
docs(readme): document unsupported TS compiler options

joelmukuthu:

- [x] ~~It's really useful if your PR relates to an outstanding issue, so please reference it in your PR, or create an explanatory one for discussion. In many cases features are absent for a reason.~~ From [this comment](https://github.com/kaisermann/svelte-preprocess/issues/98#issuecomment-572854189) on issue #98 
- [x] ~~This message body should clearly illustrate what problems it solves. If there are related issues, remember to reference them.~~ N/A
- [x] ~~Ideally, include a test that fails without this PR but passes with it. PRs will only be merged once they pass CI. (Remember to `npm run lint`!)~~ N/A

### Tests

- [x] ~~Run the tests tests with `npm test` or `yarn test`~~ not necessary, I think.

### DIFF
--- a/README.md
+++ b/README.md
@@ -457,6 +457,11 @@ svelte.preprocess(input, sveltePreprocess(options));
 
 Since `typescript` is not officially supported by `svelte` for its template language, `svelte-preprocess` only type checks code in the `<script></script>` tag.
 
+The following compiler options are not supported:
+
+- `noUnusedLocals`
+- `declarations`
+
 ### `pug`
 
 #### Template blocks
--- a/README.md
+++ b/README.md
@@ -460,6 +460,7 @@ Since `typescript` is not officially supported by `svelte` for its template lang
 The following compiler options are not supported:
 
 - `noUnusedLocals`
+- `noEmitOnError`
 - `declarations`
 
 ### `pug`


kaisermann:
Just came to my mind that we can, apart from documenting, override them in the typescript transformer: https://github.com/kaisermann/svelte-preprocess/blob/master/src/transformers/typescript.ts#L198.

What do you think?

joelmukuthu:
Hmm, I think that might cause some confusion. I think a better option is maybe to warn the user if they try to use unsupported options. I can work on that too but probably won't be till next weekend.

I'd also like to add a little more info to the README as to why they're not supported, but frankly I'm a bit confused myself. I have this setup:

- `noUnusedLocals: true` in tsconfig.json
- `typescript: { tsconfigFile: './tsconfig.json' }` in svelte-preprocess's options
- an unused local variable in a `script lang="typescript"` tag in a Svelte component

Building the project via webpack works while running tests on Jest (with ts-jest transforming TS to JS) fails with a `ParseError: Unexpected character ''` error thrown by the svelte compiler. Does svelte-preprocess already ignore/bypass the TS compilation errors?

kaisermann:
`svelte-preprocess` ignores only some import related diagnostics. Can you provide a minimal repro project?

joelmukuthu:
@kaisermann you're right. I have a setup where a lot was going on and I couldn't reproduce it in a smaller project. Thank you for your time!

PR #129, opened 2020-03-29
feat: 🎸 add `replace` transformer

kaisermann:
It came to my mind that it would useful to have an easy way to replace values inside svelte components like what `Webpack.DefinePlugin` does. I know people can use the webpack/rollup plugin instead, but adding a `markup` transformer guarantees that we're replacing values before the compilation step, which allows some interesting use cases.

### DIFF
--- a/README.md
+++ b/README.md
@@ -191,7 +191,7 @@ Which allows to write your component like:
 
 ```html
 @if(process.env.NODE_ENV !== 'development')
-  <h1>Production environment!</h1>
+<h1>Production environment!</h1>
 @endif
 ```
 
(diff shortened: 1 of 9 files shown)


ispyinternet:
@kaisermann would it be possible for this to traverse into imported files? I'm importing typescript that contains some `_CONFIG_` I want replaced. Not sure how to go about this otherwise? Also the said config is obtained via async, and with `preprocess` supporting async its kind of ideal...!

kaisermann:
Hey @ispyinternet 👋  In that case, I'd suggest using your bundler/tooling plugin for replacing values in your codebase, such as these:

- https://www.npmjs.com/package/@rollup/plugin-replace
- https://webpack.js.org/plugins/define-plugin/

`svelte-preprocess` only handles `.svelte` files. Traversing other files is out of the scope 🙏

ispyinternet:
Yh I'm just migrating to svelte-kit so got to get my head around where I can inject more complex setups into rollup / typescript as it all seems to be wrapped by svelte.config.js.

Thanks for quick reply.

Issue #148, opened 2020-05-16
Add SCSS option to explicitly set compiler


wongjn:
**Is your feature request related to a problem? Please describe.**

Due to the build process, I end up having both `node-sass` and `sass` dev dependencies installed, although I use `sass` for my code. However, this module looks for `node-sass` first which causes errors due to the fact `node-sass` does not support the new module syntax introduced in `sass`.

**Describe the solution you'd like**

Add an option into `scss` keyed with `compiler` that points to the preferred compiler:
```js
scss: {
  compiler: require('sass'),
}
```

**Describe alternatives you've considered**

None.

**How important is this feature to you?**

It would a nice-to-have so I could use module syntax in svelte components.

**Additional context**

I will shortly create a pull request with implementation.

kaisermann:
Hey @wongjn 👋 Thanks for your issue (and the PR!). If I understood your problem correctly, would switching the require order be enough to solve your problem? I think it makes sense to first look for the newest sass implementation instead of the legacy one.

wongjn:
Yes, that would solve my particular problem; I just thought it would be nice to give a choice.

kaisermann:
@wongjn I understand the idea, but I don't think many people would benefit from the extra config option. Also, I always try to prevent adding more options to keep the API as smaller as possible. I could be wrong, but for now, I think changing the order would be an easier fix! Will do in a moment

wongjn:
Sure, that's fine, thank you for your consideration!

kaisermann:
I was thinking about this one and I've changed my mind. Having the prop is more explicit and easier to understand if you want to change your scss implementation. However, I do prefer to align the prop name with `sass-loader`, that is, changing from `compiler` to `implementation`. What do you think?

wongjn:
That could be good — the prop I suggested was from the implementation in `gulp-sass`. I don't think the name matters too much so choose a name you like!

kaisermann:
@wongjn I think nowadays most people are more used to webpack instead of gulp, so let's go with `implementation` 😁!

kaisermann:
Closed in #149

PR #149, opened 2020-05-16
Add compiler option for scss

wongjn:
Initial implementation of #148


dkniffin:
@kaisermann It's funny that you're working on this right now, because I actually need this for something I just ran into 😂 . Essentially, the dart-sass implementation [supports something the libsass one doesn't](https://sass-lang.com/documentation/at-rules/mixin#passing-arguments-to-content-blocks) and I'd like to use the dart implementation for svelte instead of the libsass one. How much longer until this is merged and released?

kaisermann:
@wongjn @dkniffin Done! Released the `implementation` prop in `v3.8.0`. Thanks for all the help 😁

PR #155, opened 2020-05-24
Added support for filtering unused named imports.

### LINKED ISSUE #153: Importing external interfaces fails
If I try to import a TypeScript interface from from an external TS file, I get the following error:

`Error: '<my interface>' is not exported by <my interface file>.ts, imported by <my component>.svelte`

**To Reproduce**
Try to import an interface into a Svelte Component when using svelte-preprocess. I'm also using the TypeScript plugin as it seems that `svelte-preprocess` doesn't process plain TS files.

**Expected behavior**
I should be able to reference interfaces from external files.

- `svelte-preprocess` version: 3.7.4

- Whether your project uses Webpack or Rollup: Rollup

MaartenX:
Uses the AST to check if named imports are used in the transpiled code. If not, filters out the named import. Another option could be to use the typechecker, but then it would break when transpileOnly was true.

Fixes #153.

This way non-emitting types will never be imported so we don't get the following error:

```[!] Error: '<Interface>' is not exported by <TypeScriptFile>.ts, imported by <SvelteFile>.svelte```

### Tests

### DIFF
--- a/src/processors/typescript.ts
+++ b/src/processors/typescript.ts
@@ -1,7 +1,13 @@
 import { Options, PreprocessorGroup } from '../types';
 import { concat, parseFile } from '../utils';
 
+const markupCache: Record<string, string> = {};
+
 export default (options: Options.Typescript): PreprocessorGroup => ({
+  markup({ content, filename }: { content: string; filename: string }) {
+    markupCache[filename] = content;
+    return { code: content };
+  },
   async script(svelteFile) {
     const { default: transformer } = await import('../transformers/typescript');
     const { content, filename, lang, dependencies } = await parseFile(
@@ -10,7 +16,13 @@ export default (options: Options.Typescript): PreprocessorGroup => ({
     );
     if (lang !== 'typescript') return { code: content };
 
-    const transformed = await transformer({ content, filename, options });
+    const transformed = await transformer({
+      content,
+      filename,
+      options,
+      markup: markupCache[svelteFile.filename],
+    });
+
     return {
       ...transformed,
       dependencies: concat(dependencies, transformed.dependencies),
--- a/src/transformers/typescript.ts
+++ b/src/transformers/typescript.ts
@@ -80,9 +80,129 @@ const importTransformer: ts.TransformerFactory<ts.SourceFile> = context => {
   return node => ts.visitNode(node, visit);
 };
 
-const TS_TRANSFORMERS = {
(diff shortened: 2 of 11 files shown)

### CODE REVIEW

**Comment by MaartenX** [comment]
Just found an issue, it filters too well. If you import elements by name from a module and only use them in the markup, the types are removed also.

The best option would be to use the type checker to find out what kind of types we're importing, but the transpileOnly option would no longer work if I take that route.

File: src/transformers/typescript.ts [hunk 1]
```diff
@@ -80,8 +80,112 @@ const importTransformer: ts.TransformerFactory<ts.SourceFile> = context => {
...
+const removeNonEmittingImports: ts.TransformerFactory<ts.SourceFile> = context => {
+  function createVisitor(usages: { [name: string]: number }) {
+    const visit: ts.Visitor = node => {
+      if (ts.isImportDeclaration(node)) {
+        let importClause = node.importClause;
+        const bindings = importClause.namedBindings;
+
+        if (bindings && ts.isNamedImports(bindings)) {
+          const namedImports = bindings.elements.filter(
+            element => usages[element.name.getText()] > 0,
```
MaartenX, line 155: This filters a bit too much; need a way to find out if imports are used in the non-script part of the file.

src/processors/typescript.ts
```diff
@@ -10,7 +16,13 @@ export default (options: Options.Typescript): PreprocessorGroup => ({
     );
     if (lang !== 'typescript') return { code: content };
 
-    const transformed = await transformer({ content, filename, options });
+    const transformed = await transformer({
+      content,
+      filename,
+      markup: markupCache[svelteFile.filename],
```
dkondrad, line 22: Hi, as a suggestion, I'd move the additional parameter to the end of the param list to be consistent with your additions to other param lists earlier in the file.

MaartenX: Done.

halfnelson:
This will make the pre-processor much more useful, I run into this ALL THE TIME :)

orta:
I'm not 100% on all the problems this solves, but can recommending `import type` everywhere be an answer also? That can guarantee that it should be removed

kaisermann:
Hey @MaartenX 😁 ! First of all, thanks for your contribution! After checking this PR and #164, I've decided to go with #164 since it uses, as you suggested, the typechecker while keeping `transpileOnly` working.

PR #161, opened 2020-06-03
feat: add the :global {} rule support

illright:
This implements the proposal discussed in Svelte RFCs: https://github.com/sveltejs/rfcs/pull/22#issuecomment-638139699

```html
<style>
  .scoped-style {}
  
  :global {
    import 'global-stylesheet';
  
    .global-style {}
  }
</style>
```

This would allow for fine control of what parts of the stylesheet should be made global.

### Tests

### DIFF
--- a/src/autoProcess.ts
+++ b/src/autoProcess.ts
@@ -27,10 +27,10 @@ interface Transformers {
   postcss?: TransformerOptions<Options.Postcss>;
   coffeescript?: TransformerOptions<Options.Coffeescript>;
   pug?: TransformerOptions<Options.Pug>;
-  globalStyle?: TransformerOptions<Options.Typescript>;
-  globalRule?: TransformerOptions<Options.Typescript>;
+  globalStyle?: TransformerOptions;
+  globalRule?: TransformerOptions;
   replace?: Options.Replace;
-  [languageName: string]: TransformerOptions<any>;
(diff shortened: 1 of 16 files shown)


illright:
As recently mentioned in the discussion of the RFC, maybe going with `:global {}` would be better than `@global {}` for compliance with CSS modules. Join the discussion in the RFC if you think this PR should be changed to use the colon notation instead.

illright:
I'm not sure why the build is failing. Can't reproduce the issue on Manjaro with the same Node version that is used in the CI. Besides, I didn't even edit the Less files. Any ideas?

kaisermann:
Hey @illright 👋  First of all thanks for your pull request! I'll give it a look later today or tomorrow to see what's happening with the tests 😁

src/autoProcess.ts
```diff
@@ -270,6 +272,15 @@ export function autoPreprocess(
         map = transformed.map;
       }
 
+      const transformed = await runTransformer('globalRule', null, {
```
kaisermann, line 275: Hmmm, couldn't this thrown an error if `postcss` is not installed? We need some kind of check to run this transformer only if `postcss` is installed

kaisermann:
Done! Release in `v3.9.0` 🎉 . Thank you very much for your contribution 💯

PR #164, opened 2020-06-05
Fix/remove type imports typescript

halfnelson:
## Current problem 

Imported interfaces and types cause rollup etc to choke as they are no longer stripped. see: https://github.com/sveltejs/svelte-preprocess/issues/153

## Fix
I noticed that the typescript forces them back in as part of a transform. I add support for the detection of `import type { AInterface } ...` style imports that can be skipped. There is a more powerful version for when the transformer isn't just doing transpile only. It uses the type checker to see if the symbol represents an interface or type alias and removes it.

This is the second attempt at a fix. The existing PR ( https://github.com/sveltejs/svelte-preprocess/pull/155 ) removes the import if the value is unused in the script, but it may remove imports in error if the import is only used as prop values.

### Tests
I added a bunch of tests for this feature that now pass. The less.js test was already failing before the changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## [3.9.3](https://github.com/kaisermann/svelte-preprocess/compare/v3.9.2...v3.9.3) (2020-06-06)
+
+
+
 ## [3.9.2](https://github.com/kaisermann/svelte-preprocess/compare/v3.7.4...v3.9.2) (2020-06-06)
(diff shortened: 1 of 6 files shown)


kaisermann:
Hey @halfnelson 👋  Thanks for the great pull request! The typescript API has always been kinda cryptic too me 😅. Released in `v3.9.3` 🎉

PR #167, opened 2020-06-06
Do not break when :global is the only selector

illright:
This is... embarrassing.

Apparently, for code like this:
```sass
:global {
  /* comment */
  .cls { color: red; }
}
```

Sass and Less will produce the following output:
```css
:global {
  /* comment */
}
:global .cls { 
  color: red; 
}
```

...which leaves us with comment-only rules that only consist of the `:global` selector, which gets lovingly transpiled into an empty string, producing things like `<style>{/* comment */}</style>`. This is to remedy that.

### Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,13 @@
+## [3.9.4](https://github.com/kaisermann/svelte-preprocess/compare/v3.9.2...v3.9.4) (2020-06-06)
+
+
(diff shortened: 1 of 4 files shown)


kaisermann:
😆  nice catch! I've modified it a bit to support multiple selectors like `:global, div {}` (who knows...). Released in `v3.9.4` 🎉

PR #168, opened 2020-06-06
Make the selector parsing smarter 

illright:
Previously, the selectors were split by a single literal space, which is rather naive. This PR allows for all kinds of whitespace, combinators like `>`, `+`, `~`, as well as spaces that aren't combinators (`[attr="with spaces"]`)

### Tests


illright:
Oh, right, I forgot the comma combinator

illright:
Also, I'm not sure why, but when I run `yarn lint` locally, it doesn't detect some of the `prettier` errors that are detected by the CI. Am I running the wrong command?

kaisermann:
Oh, that was indeed naive 👀  Thanks 🙌

PR #173, opened 2020-06-09
Fix the quotes in the release script

illright:
In Bash, single quotes prevent variable substitution, which results in all of the releases having `Release v$npm_package_version` as their tag message. This is to fix that.

Also, it would be nice to incorporate the CHANGELOG.md into those messages as well, but I don't know how would one do that.


kaisermann:
I've been wanting to rethink the publish/release flow for a while. I'll merge this for now and start looking for alternatives. I don't think automated changelogs are THAT good anymore 😓

PR #210, opened 2020-08-04
Correct a wrong variable name in the example code

jakubferenc:
The variable name in the code given as an example was wrong. It should be "name" instead of "world": {world} => {name}


kaisermann:
Thanks 🎉 !

PR #223, opened 2020-08-14
Update preprocessing.md

freethebit:
Add example and description for SCSS and SASS on how to prepend data.


kaisermann:
Thanks! 🎉  I've removed the example as it's already described how to do it [here](https://github.com/sveltejs/svelte-preprocess/blob/master/docs/getting-started.md#32-prepending-content).

Issue #255, opened 2020-09-22
Do preprocess without bundling


pd4d10:
**Is your feature request related to a problem? Please describe.**

When using svelte to develop a library, it is recommended to ship source code (*.svelte) to npm and add a `svelte` field to `package.json`([from this document](https://github.com/sveltejs/rollup-plugin-svelte#pkgsvelte)). If this library is used by someone, the rollup plugin would take these source code and recompile them.

The problem is, if some preprocessors are used in the library development (for example TypeScript), all users of this lib have to add these preprocessors to their app/library project, because rollup plugin would take the source code and recompile them.

**Describe the solution you'd like**

Provide a way to do preprocess without bundling. Take TypeScript as an example, input is `<script lang="ts">/* TS code */</script>` and output is `<script>/* JS code */</script>`

**How important is this feature to you?**

We encountered this problem when developing [ByteMD](https://github.com/bytedance/bytemd), which is a hackable Markdown editor built with Svelte.

kaisermann:
Hey, @pd4d10 👋  This feature is unrelated to `svelte-preprocess` since you can achieve what you want by calling`svelte-preprocess` in some node script. However, if more people show an interest in an easy way to just preprocess their components to some directory, we can create a simple CLI for that.

pd4d10:
OK, I will do some digging and send a PR if possible

PR #263, opened 2020-10-05
WIP - Feat/cli

### LINKED ISSUE #256: CLI for preprocessing svelte files
Tries to tackle https://github.com/sveltejs/component-template/issues/8 and #255.

A tiny CLI (possibly using `sade`) to preprocess invalid `.svelte` files, which are files written in languages other than JS, CSS, and HTML, into valid `.svelte` files. This is especially useful for library authors that want to write their components using some custom language such as SASS or TypeScript.

```
$ npx svelte-preprocess --src=/ --out=dist/
```

https://github.com/sveltejs/component-template/pull/31

kaisermann:
Closes #256, #255 

Related to: https://github.com/sveltejs/component-template/pull/31 and https://github.com/sveltejs/component-template/issues/8

Opening it as a draft and with no implementation. I'm currently studying the available routes we can take here.

### DIFF
--- a/package.json
+++ b/package.json
@@ -94,6 +94,7 @@
     "@types/pug": "^2.0.4",
     "@types/sass": "^1.16.0",
     "detect-indent": "^6.0.0",
+    "sade": "^1.7.4",
     "strip-indent": "^3.0.0"
   },
   "peerDependencies": {
--- a/src/cli/index.ts
+++ b/src/cli/index.ts
@@ -1,0 +1,9 @@
+import sade from 'sade';
+
+import pkg from '../../package.json';
+
+const prog = sade('svelte-preprocess');
+
+prog.version(pkg.version);
+
+prog.parse(process.argv);
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -9,7 +9,8 @@
     "lib": ["es2018"],
     "declaration": true,
     "types": ["jest", "node"],
-    "outDir": "dist"
+    "outDir": "dist",
+    "resolveJsonModule": true
   },
   "include": ["**/*.ts"],
   "exclude": ["node_modules/**/*", "dist"]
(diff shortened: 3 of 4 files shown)


lukeed:
Anything I can help with?

kaisermann:
@lukeed 🤔 We need to decide what route we should take here. Initially, I was thinking of a very simple CLI that would just preprocess something(s) from `--src` and output it(them) to `--output`

```sh
$ svelte-preprocess --src src/*.svelte --out dist/
```

This can be enough for some (most?) cases,  but as @dummdidumm noted [here](https://github.com/sveltejs/component-template/pull/31#pullrequestreview-483990582), a component can import some, i.e `ts` module and that obviously wouldn't work from the consumer perspective. Bundling/transpiling these modules don't seem to be the correct way forward to me. Maybe it's just a matter of limiting the scope of what the CLI does.

lukeed:
I would 100% just worry about compiling standalone files. I'm pretty sure the old `svelte` CLI did this too. 
Otherwise, yeah, you definitely encroach on being a full bundler/Rollup wrapper.

If output is targeting ESM, then it's easy to position this as a "prepare your component for bundling" step, since the output is pure JS, CSS, HTML

kaisermann:
@lukeed I like the idea of positioning the CLI as a bundling "preparation" step, sparing us from dealing with all the minutia of js bundling. However, if we follow this path we should create at least some examples, one simple and another a bit more complex. Maybe the component-template can be used as the example itself, not sure. 

Anyways, do you (or anyone else) have thoughts about the CLI API? For now, I was thinking of:

`--input` | `--src` - **required**
Glob(?) of svelte files to be preprocessed

`--output` - **required**
Output directory of preprocessed components. 
I thought of following the same file structure found in `--input`, but if the components reside in an `src/components`, how could one make the components be output to the root of a `dist/` directory?

`--config` - **optional**
Path to a `svelte.config.js`. Defaults to `process.cwd()/svelte.config.js`

firefish5000:
Exert of relative code from `@tdi/svcli` for reference. <https://gist.github.com/firefish5000/372ec181a5c47ac5190358639182310f>
There is not to much special going on compared to the component-template or the code I stole from here somewhere. It also copies files based off of a glob (which is not necessary by any means). But if this only deals with svelte files we need to be sure to mention how to fix `*.js` imports (along with custom imports which need their own preprocessors).

Note the main thing it adds that I think is necessary is an `extensions` field which allows custom extensions like `.svx`. A input glob would work just as well.

I am also of the opinion that this should only deal with svelte files (ideally with customizable extensions) as anything more than that quickly turns it into a task runner like gulp.

Issue #277, opened 2020-11-06
Please add an import resolver for tilde (~) node_modules imports


GitGangGuy:
I know that svelte-preprocess adds `node_modules` to the scss/sass `includePaths` automatically, but this will only resolve import/use statements in the form of `module/...`, not `~module/...`, which has become a standard for importing modules in scss. Even VSCode will highlight and link paths relative to `node_modules` when prefixing them with a tilde (a path without one will however not be resolved correctly).
Would it be possible to treat tilde-prefixed paths as relative to `node_modules` as well, like a lot of webpack plugins have been doing already?

kaisermann:
Hey @GitGangGuy 👋 

I think it's possible to add a custom `importer` that just removes the initial `~` character and leave the rest to the `includePaths` resolution. Will come back to you.

GitGangGuy:
@kaisermann Thank you for addressing this!

kaisermann:
It seems sass doesn't pass an importer returned value through the `includePaths` and it expects an absolute path. This complicates it a little bit more. For now, you could try using [`node-sass-tilde-importer`](https://github.com/matthewdavidson/node-sass-tilde-importer) and see if it works for you!

GitGangGuy:
@kaisermann I use dart `sass`, so `node-sass` extensions aren't an option. Thanks for getting it working anyway!

kaisermann:
@GitGangGuy AFIK both `sass` and `node-sass` importers have the same API. It's just a function.

GitGangGuy:
The PR was merged, why is this issue still open? ^^
I'll go ahead and close it

PR #278, opened 2020-11-07
feat: 🎸 add scss support for tilde (~) imports

kaisermann:
✅ Closes: #277

edit 1:
~Just gotta do a test for it 😁~

edit 2:
~Tests are failing now, not sure why. See https://github.com/sass/node-sass-middleware/issues/104~ Can't use the `done` for `renderSync` runs. 

Current challenge: 
it seems the sass compiler doesn't pass an importer's returned value through the `includePaths` paths, so it needs to return an absolute path. We use the `url` and `prev` (source) arguments to resolve it ~, but, for some reason, the `prev` value is coming as `stdin` instead of the importer path.~. Needed to pass `file` alongside the rest of the options.

edit 3:
Need to rewrite that importer, the logic is wrong. It will stop on the first node_modules found, even if there's no package with the specified path.

References:
https://github.com/matthewdavidson/node-sass-tilde-importer/blob/master/index.js
https://sass-lang.com/documentation/js-api#importer

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -7,6 +7,7 @@
   "rules": {
     "@typescript-eslint/no-explicit-any": "off",
     "@typescript-eslint/ban-types": "off",
-    "no-console": "off"
+    "no-console": "off",
+    "line-comment-position": "off"
   }
 }
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,16 +14,16 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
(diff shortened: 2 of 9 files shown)


kaisermann:
Not sure why the windows test is failing 😢

GitGangGuy:
@kaisermann Hi, do you know why the windows test isn't working? I didn't even know certain NodeJS modules/features are platform-dependent  😮

kaisermann:
@GitGangGuy Nope, no idea. The file is there 🤷 However, it may work if I read it manually and use the `{ content }` return API instead of `{ file }`.

PR #289, opened 2020-12-07
Fix VS Code warning

benmccann:
This fixes a VS Code warning I was getting:

> 'code' is already declared in the upper scope.

### DIFF
--- a/src/transformers/coffeescript.ts
+++ b/src/transformers/coffeescript.ts
@@ -25,9 +25,7 @@ const transformer: Transformer<Options.Coffeescript> = ({
     return { code, map };
   }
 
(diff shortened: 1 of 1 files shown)


kaisermann:
Hmmm, where are you seeing this warning? I'm not getting it over here

benmccann:
Screenshot from 2020-12-07 10-53-11

kaisermann:
Oh, I didn't know about the SonarLint extension, sweet 😁 

I'm all up against variable shadowing, but I don't see the problem here. The `code` variable from the outer scope cannot be used before its declaration 🤔, but oh well 🤷

kaisermann (approved):

PR #291, opened 2020-12-11
Sync preprocess

ehrencrona:
This is a proposal for allowing synchronous preprocessing in order to solve sveltejs/kit#19 . It allows preprocessors to export synchronous versions of their processing functions. This allows the preprocessor to be used with e.g. `require.extensions`.

It goes together with sveltejs/svelte#5770

It's still a draft; tests are missing and the code could be refactored further. I'd like feedback as to whether this is a good idea before proceeding.

Getting TypeScript to play ball with functions that are sometimes sync and sometimes async was a bit messy, which IMO is the main argument against this solution.

### DIFF
--- a/src/autoProcess.ts
+++ b/src/autoProcess.ts
@@ -1,12 +1,8 @@
 /* eslint-disable @typescript-eslint/naming-convention */
-import {
-  PreprocessorOptions,
-  SyncPreprocessor,
-} from 'svelte/types/compiler/preprocess';
+import { PreprocessorOptions } from 'svelte/types/compiler/preprocess';
 
 import {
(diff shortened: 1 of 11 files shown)


kaisermann:
Hey @ehrencrona 👋  First of all, thanks for the effort in trying to resolve this async problem. However, I'm not convinced that this is something that should be handled on our side. The current PR adds a fair bit of complexity and it's exactly a scalable solution to the problem. IMO using a `deasync`-like lib or starting a new node process, albeit not the ideal solution, is a better way forward for now.

ehrencrona:
Closing due to lacking traction.

Issue #286, opened 2020-12-05
Sourcemaps offset


Maggi64:
**Describe the bug**
In my current project and in the example repo below the sourcemaps are offset by a few lines. 
This seems to be related to the style tag:
![image](https://user-images.githubusercontent.com/20517052/101258090-70730780-3720-11eb-9788-ba7b6d73bc67.png)

**To Reproduce**  
- clone repo (https://github.com/Maggi64/sapper-sourcemap-offset)
- `npm run dev`
- look at chrome dev tools

**Expected behavior**
Correctly linked sourcemaps. Otherwise debugging issues becomes quite hard.

**Information about your project:**

- Your browser and the version: (Chrome 87)

- Your operating system: (Windows 10)

- Rollup & svelte-preprocess are updated to the newest versions (see repo package.json)

- All other packages used are also on the newest version 

I hope this is the correct place for this issue. Help and suggestions where that problem exactly comes from are appreciated.

Edit: Even without scss preprocessing this issue still persists. Adjusted example repo.

benmccann:
I'm guessing this is a bug in Svelte: https://github.com/sveltejs/svelte/pull/5584#issuecomment-733811305. @milahu I know you had done some investigation into this. Not sure if it's something you'll be able to take a look at, but if so I'd be happy to help review

milahu:
thanks for the bug : )

a minimal reproduction is: add this to `app.svelte`
```js
throw new Error('this should be line 10');
```
and then look at the JS console
which traces the exception back to `app.svelte` line 11, for example

triggered by using the `svelte-preprocess` plugin
so either blame `svelte-preprocess` or `svelte`
im adding a test to svelte to find out more ..

@dummdidumm do you know more?
ben said you saw the off-by-one bug too

dummdidumm:
No I don't know more, but I think it's inside Svelte because inside language-tools we are able to map the warnings correctly.

milahu:
small bug in svelte-preprocess:

the regex for `<template>.*</template>` is too greedy
so the input `<template>T1</template><template>T2</template>`
is wrongly replaced with `T1</template><template>T2`

these work as expected in svelte:
```js
	async function preprocess_tag_content(tag_name: 'style' | 'script', preprocessor: Preprocessor) {
		const get_location = getLocator(source);
		const tag_regex = tag_name === 'style'
[...]

PR #297, opened 2021-01-06
sass: remove sourceMappingURL from result.code

milahu:
related to issue #286 and PR https://github.com/sveltejs/svelte/pull/5854

sourcemaps should be returned either in result.map or in result.code but not in both

keeping the `/*# sourceMappingURL=component.svelte.map */` magic comment 
~~makes the browser display wrong mappings~~
\*can\* make the browser display wrong mappings, if the sourcemap file exists (`component.svelte.map`)

~~(there is another bug with sourcemaps, so mappings are still wrong in the browser)~~
fixed by https://github.com/sveltejs/rollup-plugin-svelte/issues/175

### DIFF
--- a/src/transformers/scss.ts
+++ b/src/transformers/scss.ts
@@ -68,6 +68,7 @@ const transformer: Transformer<Options.Sass> = async ({
     ...options,
     includePaths: getIncludePaths(filename, options.includePaths),
     outFile: `${filename}.css`,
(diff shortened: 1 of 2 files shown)


benmccann (approved):
lgtm. Is there a test we can add for this?

kaisermann (approved):

kaisermann:
Ni-ni-ni-nice! Thanks! Released in v4.6.5 🎉

PR #299, opened 2021-01-18
fix typescript: translate options.sourceMap to options.compilerOptions.sourceMap (#286)

milahu:
trivial : )

edit: less trivial on second sight
in the old version, `sveltePreprocess({ sourceMap: true })` 
was translated to `transformer({ content, filename, options: { sourceMap: true } })`
but the typescript `transformer` function expects compiler-options in `options.compilerOptions`
so the correct translation is `transformer({ content, filename, options: { compilerOptions: { sourceMap: true } } })`

### DIFF
--- a/src/autoProcess.ts
+++ b/src/autoProcess.ts
@@ -131,9 +131,9 @@ export function sveltePreprocess(
       const pathParts = propPath.split('.');
       let parentObj = opts;
       let i;
-      for (i = 0; i < (pathParts.length - 1); i++) {
+      for (i = 0; i < pathParts.length - 1; i++) {
         const propName = pathParts[i];
-        if (typeof parentObj[propName] != 'object') parentObj[propName] = {};
+        if (typeof parentObj[propName] !== 'object') parentObj[propName] = {};
         parentObj = parentObj[propName];
       }
       const propName = pathParts[i];
--- a/src/autoProcess.ts
+++ b/src/autoProcess.ts
@@ -127,9 +127,17 @@ export function sveltePreprocess(
     }
 
     if (sourceMap && name in SOURCE_MAP_PROP_MAP) {
-      const [propName, value] = SOURCE_MAP_PROP_MAP[name];
-
-      opts[propName] = value;
+      const [propPath, value] = SOURCE_MAP_PROP_MAP[name];
+      const pathParts = propPath.split('.');
+      let parentObj = opts;
+      let i;
+      for (i = 0; i < (pathParts.length - 1); i++) {
+        const propName = pathParts[i];
+        if (typeof parentObj[propName] != 'object') parentObj[propName] = {};
+        parentObj = parentObj[propName];
+      }
+      const propName = pathParts[i];
+      parentObj[propName] = value;
     }
 
     return opts;
--- a/src/autoProcess.ts
+++ b/src/autoProcess.ts
@@ -7,7 +7,7 @@ import type {
   Transformers,
   Options,
 } from './types';
-import { hasDepInstalled, concat } from './modules/utils';
+import { hasDepInstalled, concat, setProp } from './modules/utils';
 import { getTagInfo } from './modules/tagInfo';
 import {
   addLanguageAlias,
@@ -127,23 +127,7 @@ export function sveltePreprocess(
     }
 
     if (sourceMap && name in SOURCE_MAP_PROP_MAP) {
-      const [propPath, value] = SOURCE_MAP_PROP_MAP[name];
-      const pathParts = propPath.split('.');
-      let parentObj = opts;
-
-      for (let i = 0; i < pathParts.length - 1; i++) {
-        const propName = pathParts[i];
-
-        if (typeof parentObj[propName] !== 'object') {
-          parentObj[propName] = {};
-        }
-
-        parentObj = parentObj[propName];
-      }
-
-      const propName = pathParts[pathParts.length - 1];
-
-      parentObj[propName] = value;
+      setProp(opts, ...SOURCE_MAP_PROP_MAP[name]);
     }
 
     return opts;
--- a/src/autoProcess.ts
+++ b/src/autoProcess.ts
@@ -130,13 +130,19 @@ export function sveltePreprocess(
       const [propPath, value] = SOURCE_MAP_PROP_MAP[name];
       const pathParts = propPath.split('.');
       let parentObj = opts;
-      let i;
-      for (i = 0; i < pathParts.length - 1; i++) {
+
+      for (let i = 0; i < pathParts.length - 1; i++) {
         const propName = pathParts[i];
-        if (typeof parentObj[propName] !== 'object') parentObj[propName] = {};
+
+        if (typeof parentObj[propName] !== 'object') {
+          parentObj[propName] = {};
+        }
+
         parentObj = parentObj[propName];
       }
-      const propName = pathParts[i];
+
+      const propName = pathParts[pathParts.length - 1];
+
       parentObj[propName] = value;
     }
 
--- a/src/modules/utils.ts
+++ b/src/modules/utils.ts
@@ -90,3 +90,21 @@ export function findUp({ what, from }) {
 
   return null;
 }
+
+// set deep property in object
+// args = array of prop-names, last arg = value
+export function setProp(obj, ...args) {
+  let i = 0;
+
+  for (; i < args.length - 2; i++) {
+    const key = args[i];
+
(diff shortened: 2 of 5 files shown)


dummdidumm:
Not sure if this is the correct fix, since other languages do not output sourcemaps by default. I think instead it would be better if the `sourceMap` option that you can pass into `svelte-preprocess` is used.

line ~105
```ts
    ...(convertedCompilerOptions as CompilerOptions),
    importsNotUsedAsValues: ts.ImportsNotUsedAsValues.Error,
    allowNonTsExtensions: true,
    sourceMap: !!options.sourceMap // <<-- new
  };
```

milahu:
thanks, fixed

dummdidumm (approved):

dummdidumm:
My bad, I thought `options` is what one can pass in into `sveltePreprocess({..})`. So I guess some more digging is needed on how to get that option in there.

milahu:
looks like a bug in `getTransformerOptions`
```ts
    if (sourceMap && name in SOURCE_MAP_PROP_MAP) {
      const [propName, value] = SOURCE_MAP_PROP_MAP[name];

      opts[propName] = value;
    }
```

<details>
<summary>long story</summary>

```ts
// svelte-preprocess/src/transformers/typescript.ts
const transformer: Transformer<Options.Typescript> = ({
  content,
  filename,
  options = {},
}) => {
```

```ts
// svelte-preprocess/src/autoProcess.ts
export const transform = async (
  name: string,
  options: TransformerOptions,
  { content, map, filename, attributes }: TransformerArgs<any>,
): Promise<Processed> => {
  if (options === false) {
    return { code: content };
  }

  if (typeof options === 'function') {
    return options({ content, map, filename, attributes });
  }

  // todo: maybe add a try-catch here looking for module-not-found errors
  const { transformer } = await import(`./transformers/${name}`);

  return transformer({
    content,
    filename,
    map,
    attributes,
    options: typeof options === 'boolean' ? null : options,
  });
```
```ts
// svelte-preprocess/src/autoProcess.ts
    const transformerOptions = getTransformerOptions(lang, alias);
// ...
    const transformed = await transform(lang, transformerOptions, {
      content,
      filename,
      attributes,
    });
```

```ts
// svelte-preprocess/src/autoProcess.ts
  const getTransformerOptions = (
    name: string,
    alias?: string,
  ): TransformerOptions<unknown> => {
    const { [name]: nameOpts, [alias]: aliasOpts } = transformers;

    if (typeof aliasOpts === 'function') return aliasOpts;
    if (typeof nameOpts === 'function') return nameOpts;
    if (aliasOpts === false || nameOpts === false) return false;

    const opts: Record<string, any> = {};

    if (typeof nameOpts === 'object') {
      Object.assign(opts, nameOpts);
    }

    Object.assign(opts, getLanguageDefaults(name), getLanguageDefaults(alias));

    if (name !== alias && typeof aliasOpts === 'object') {
      Object.assign(opts, aliasOpts);
    }

    if (sourceMap && name in SOURCE_MAP_PROP_MAP) {
      const [propName, value] = SOURCE_MAP_PROP_MAP[name];

      opts[propName] = value;
    }

    return opts;
  };
```

```ts
// svelte-preprocess/src/autoProcess.ts
export function sveltePreprocess(
  {
    aliases,
    markupTagName = 'template',
    preserve = [],
    defaults,
    sourceMap = process?.env?.NODE_ENV === 'development' ?? false,
    ...rest
  } = {} as AutoPreprocessOptions,
): AutoPreprocessGroup {
  const defaultLanguages = Object.freeze({
    markup: 'html',
    style: 'css',
    script: 'javascript',
    ...defaults,
  });

  const transformers = rest as Transformers;
```
</details>

dummdidumm (approved):

dummdidumm:
This will also fix [this issue](https://github.com/sveltejs/language-tools/issues/489)

kaisermann:
Hey, @milahu 👋  thanks for the contribution! I'm not sure if I'm a big fan of duplicating almost the same code. I think we could extract it to a function or use lukeed's [`dset`](https://github.com/lukeed/dset/blob/master/src/index.js) (which is very tiny, as usual). What do you think? cc @dummdidumm

milahu:
hey dawg, i heard you like small functions, so me and my man `terser`, we made you a 124 byte function

```js
function setProp(t,...o){let e=0;
for(;e<o.length-2;e++){const p=o[e];"object"!=typeof t[p]&&(t[p]={}),t=t[p]}
t[o[e]]=o[e+1]}
```

<details>
<summary>verbose</summary>

```js
// args = array of prop-names, last arg = value
function setProp(obj, ...args) {
  let i = 0;

  for (; i < args.length - 2; i++) {
    const key = args[i];

    if (typeof obj[key] !== 'object') {
      obj[key] = {};
    }

    obj = obj[key];
  }

  //const key = args[i];
  //const val = args[i + 1];
  //obj[key] = val;
  
  obj[args[i]] = args[i + 1];
}

const SOURCE_MAP_PROP_MAP = {
  typescript: ['compilerOptions', 'sourceMap', true],
};
const opts = {};
const name = 'typescript';
setProp(opts, ...SOURCE_MAP_PROP_MAP[name]);
console.dir(opts);
// --> { compilerOptions: { sourceMap: true } }
```

</details>

move to `svelte-preprocess/src/modules/utils.ts`?

kaisermann:
😆 If we're going to maintain it, I tend to prefer the verbose version heh. `utils.ts` is fine!

src/modules/utils.ts
```diff
@@ -90,3 +90,21 @@ export function findUp({ what, from }) {
 
   return null;
 }
+
+// set deep property in object
+// args = array of prop-names, last arg = value
+export function setProp(obj, ...args) {
```
kaisermann, line 96: I see what you did here, but I think a signature of setProp(obj, path, value) would be easier to grasp. What do you think?

src/autoProcess.ts
```diff
@@ -127,9 +127,7 @@ export function sveltePreprocess(
     }
 
     if (sourceMap && name in SOURCE_MAP_PROP_MAP) {
-      const [propName, value] = SOURCE_MAP_PROP_MAP[name];
-
-      opts[propName] = value;
+      setProp(opts, ...SOURCE_MAP_PROP_MAP[name]);
```
kaisermann, line 130: Nice! However, I think something like this would be easier to understand:

```suggestion
      const [propPath, value] = SOURCE_MAP_PROP_MAP[name];
      setProp(opts, propPath, value);
```

kaisermann (approved):

kaisermann (changes requested):
accidentally approved it, heh sorry

src/modules/utils.ts
milahu: > I see what you did here

i hope others will see it too ; )

alternative: store the key-path in a separate array

```js
SOURCE_MAP_PROP_MAP = {
  typescript: [ ['compilerOptions', 'sourceMap'], true ],
  scss: [ ['sourceMap'], true ],
};

export function setProp(obj, keyList, val) {
  let i = 0;

  for (; i < keyList.length - 1; i++) {
    const key = keyList[i];

    if (typeof obj[key] !== 'object') {
      obj[key] = {};
    }

    obj = obj[key];
  }

  obj[keyList[i]] = val;
}

const name = 'typescript';
const opts = {};

setProp(opts, ...SOURCE_MAP_PROP_MAP[name]);

if (explicitAndVerboseCodingLevel == 100) {
  const [propPath, value] = SOURCE_MAP_PROP_MAP[name];
  setProp(opts, propPath, value);
}
```
benefit: the function signature is self-explanatory, and needs no comment
i avoided the `Array.isArray(keyList)` vs `typeof keyList == 'string'` switch, to reduce overhead

kaisermann: I like that better. Separating the path from the value eases the understandability

src/autoProcess.ts
milahu: so verbose : (

i say its clear, when you see
```js
setProp(opts, ...SOURCE_MAP_PROP_MAP[name]);
```
and

```js
export function setProp(obj, keyList, val) {
```
and

```js
export const SOURCE_MAP_PROP_MAP: Record<string, [string[], any]> = {
  babel: [['sourceMaps'], true],
  typescript: [['compilerOptions', 'sourceMap'], true],
```
that 1+1+1==3

kaisermann (approved):

kaisermann:
Thanks! Should be up in `v4.6.2`

PR #302, opened 2021-01-29
feat: 🎸 support extends prop in tsconfig

### LINKED ISSUE #300: Support "extends" tsconfig directive
svelte-preprocess seems to not follows the extends directive provided in a tsconfig.

Using `typescript.parseJsonConfigFileContent` should do the trick.

It's not blocking, since i could implement my custom tsconfig file reader, but it would be nice to have this feature built-in.

kaisermann:
Closes #300

### DIFF
--- a/src/transformers/typescript.ts
+++ b/src/transformers/typescript.ts
@@ -1,4 +1,4 @@
-import { dirname } from 'path';
+import { dirname, join } from 'path';
 
 import ts from 'typescript';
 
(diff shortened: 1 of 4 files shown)


src/transformers/typescript.ts
```diff
@@ -87,6 +65,21 @@ const transformer: Transformer<Options.Typescript> = ({
...
+          dirname(tsconfigFile),
+          config.extends,
+        );
+
+        Object.assign(
+          compilerOptionsJSON,
+          // eslint-disable-next-line @typescript-eslint/no-var-requires, node/global-require, @typescript-eslint/no-require-imports
+          require(parentConfigPath).compilerOptions,
+        );
+      }
```
kaisermann, line 82: This is the important bit

src/transformers/typescript.ts
```diff
@@ -102,11 +95,45 @@ const transformer: Transformer<Options.Typescript> = ({
...
+        node.importClause,
+        node.moduleSpecifier,
+      );
+    }
+
+    return ts.visitEachChild(node, (child) => visit(child), context);
+  };
+
+  return (node) => ts.visitNode(node, visit);
+};
```
kaisermann, line 128: Just moved this

src/transformers/typescript.ts
```diff
@@ -87,6 +65,21 @@ const transformer: Transformer<Options.Typescript> = ({
       }
 
       Object.assign(compilerOptionsJSON, config.compilerOptions);
+
+      // Support extending another tsconfig
+      if (typeof config.extends === 'string') {
+        const parentConfigPath = join(
+          process.cwd(),
```
kaisermann, line 72: I'm not sure about using cwd as base here, but we need an absolute path here.

PR #304, opened 2021-01-29
Update preprocessing documentation

benmccann:
* Expanded the example of using multiple preprocessors. It was difficult to compare to the example of using only one preprocessor because so much was different beyond just adding a second preprocessor and providing them via an array
* Added `babel` which was missing from the list of preprocessors
* Added a missing comma in the list of preprocessors
* Removed the example of `sveltePreprocess` because there was already an example of how it's called, so it didn't add much and was a little confusing to have multiple examples to have to read
* Showed where options are passed with `{ ... }`. The example above did the same, so this made it easier to compare
* Mentioned up-front that you need `postcss` for the global rule. Turned the footnotes into bullets - there was no corresponding number in the text above, so I didn't end up reading any of the footnotes as a result

### DIFF
--- a/docs/preprocessing.md
+++ b/docs/preprocessing.md
@@ -48,17 +48,19 @@ export default {
 As `svelte-preprocess` is just a Svelte preprocessor like any other, it's also possible to use it alongside other preprocessors:
 
 ```js
-import preprocess from 'svelte-preprocess'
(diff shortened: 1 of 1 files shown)


kaisermann (approved):
great! thanks for this 🙏

PR #307, opened 2021-02-03
Bump postcss-load-config

benmccann:
Closes https://github.com/sveltejs/svelte-preprocess/issues/306

(the lint failure is unrelated to this PR)


dummdidumm:
Could this be `2 | 3` (not sure about the exact syntax)? After all, it's not required for people to use version 3, only if they want cjs support.

tanhauhau (approved):

benmccann:
Is there a compelling reason to want people to be able to use v2? Normally I see us supporting multiple versions for things like Typescript or Postcss where there may be some difficulty in upgrading, but this one seems so easy to upgrade

dummdidumm:
`peerDependencies` means "user, please install this on your own", and `svelte-preprocess` will work perfectly fine with `postcss-load-config` 2.x . Only if you want to use SvelteKit, you need version 3.

benmccann:
Yeah, but I don't know why anyone should want to stay on version 2

dummdidumm:
Maybe someone uses a PostCSS plugin that needs 2.x behavior (don't know what changed for 3.x). Making `peerDependencies` accept more versions better describes the "I don't really care which version you use"-notion of `svelte-preprocess`. It's SvelteKit which needs 3.x to work, not `svelte-preprocess`.

benmccann:
It makes sense to me to support multiple postcss versions. This isn't actually the postcss version though, which is specified as `^7 || ^8`. It's just a library that reads your `postcss.config.js` file from disk if I understand correctly. The only breaking change is that it requires Node 10, which we already require anyway: https://github.com/postcss/postcss-load-config/blob/master/CHANGELOG.md#breaking-changes

dummdidumm (approved):
If that's the only breaking change then I'm okay with that.

package.json
```diff
@@ -100,7 +100,7 @@
     "coffeescript": "^2.5.1",
     "less": "^3.11.3",
     "postcss": "^7 || ^8",
-    "postcss-load-config": "^2.1.0",
+    "postcss-load-config": "^3.0.0",
```
kaisermann, line 103:
```suggestion
    "postcss-load-config": "^2.1.0 || ^3.0.0",
```

kaisermann (approved):

kaisermann:
Thanks! I've added v2 to the peerDep as well (as @dummdidumm rightly suggested). Released in 4.6.6 :tada:

PR #319, opened 2021-03-02
fix destructuring errors in esbuild example in preprocessing.md

SpicyRicecaker:
In the `preprocessing.md` doc there seems to be a destructuring error. 

```javascript
const { js: code } //... js doesn't exist
return { code };
```
Using the code above makes it so that the compiler throws errors. The latest esbuild [transform api](https://esbuild.github.io/api/#transform-api) doesn't return an object with the `js` property but instead has the `code` property. 

The small changes below fixes the error and actually mirrors some of the code in the *custom preprocessors* section slightly above the esbuild example.
```javascript
const { code, map } = //...
```

### DIFF
--- a/docs/preprocessing.md
+++ b/docs/preprocessing.md
@@ -155,12 +155,11 @@ export default {
   plugins: [
     svelte({
       preprocess: sveltePreprocess({
-        typescript({ content, filename }) {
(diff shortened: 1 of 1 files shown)


kaisermann (approved):
Nice! Thanks :pray:

PR #326, opened 2021-03-03
Docs: getting-started.md - Change examples to use {name} variable

danmarshall:
I'm following the final example on this page, which uses {name} in the template. I don't think there's a `world` variable, but I could be wrong 🌎


kaisermann (approved):
👀

PR #328, opened 2021-03-11
(feat) Support tsconfig extends

### LINKED ISSUE #300: Support "extends" tsconfig directive

dummdidumm:
Fixes #300

I was unsure about the boolean/string behavior of the options, especially because it is mentioned nowhere in the docs that boolean is a valid option.

Also, if I understand this correctly, by default the tsconfig.json is not picked up if I do `sveltePreprocess()`?

### Tests

### DIFF
--- a/src/transformers/typescript.ts
+++ b/src/transformers/typescript.ts
@@ -70,6 +70,7 @@ export function loadTsconfig(
   let tsconfigFile =
     tsOptions.tsconfigFile ||
     ts.findConfigFile(fileDirectory, ts.sys.fileExists);
+
   tsconfigFile = isAbsolute(tsconfigFile)
     ? tsconfigFile
     : join(basePath, tsconfigFile);
@@ -92,6 +93,7 @@ export function loadTsconfig(
     compilerOptionsJSON,
     tsconfigFile,
   );
+
   // Filter out "no files found error"
   errors = errors.filter((d) => d.code !== 18003);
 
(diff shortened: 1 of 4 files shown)


kaisermann:
Thanks for this! Will give it a look later

> Also, if I understand this correctly, by default the tsconfig.json is not picked up if I do sveltePreprocess()?

It is! If `tsconfigFile` is not `false` (the boolean that is not documented 😳 ), we start looking for the `tsconfig.json` in the `dirname(filename)`.

https://github.com/sveltejs/svelte-preprocess/blob/main/src/transformers/typescript.ts#L69

src/transformers/typescript.ts
```diff
@@ -53,6 +53,53 @@ const importTransformer: ts.TransformerFactory<ts.SourceFile> = (context) => {
...
+  let basePath = process.cwd();
+
+  const fileDirectory = (tsOptions.tsconfigDirectory ||
+    dirname(filename)) as string;
+
+  let tsconfigFile =
+    tsOptions.tsconfigFile ||
+    ts.findConfigFile(fileDirectory, ts.sys.fileExists);
+
+  tsconfigFile = isAbsolute(tsconfigFile)
```
dummdidumm, line 74: converting to absolute path here because else `parseJsonConfigFileContent` doesn't seem to work

kaisermann (approved):

kaisermann:
🎉  neat 🎉

PR #337, opened 2021-03-29
fix: issues/336: typescript: importing component from pnpm monorepo p…

btakita:
fixes https://github.com/sveltejs/svelte-preprocess/issues/336

### DIFF
--- a/src/transformers/typescript.ts
+++ b/src/transformers/typescript.ts
@@ -143,15 +143,16 @@ const transformer: Transformer<Options.Typescript> = ({
     );
   }
 
+  const absoluteFilename = isAbsolute(filename)
+    ? filename
+    : resolve(basePath, filename);
+
   const {
     outputText: code,
     sourceMapText: map,
     diagnostics,
   } = ts.transpileModule(content, {
-    fileName:
-      !isAbsolute(filename)
-        ? resolve(join(basePath, filename))
-        : filename,
+    fileName: absoluteFilename,
     compilerOptions,
     reportDiagnostics: options.reportDiagnostics !== false,
     transformers: {
--- a/src/transformers/typescript.ts
+++ b/src/transformers/typescript.ts
@@ -149,8 +149,8 @@ const transformer: Transformer<Options.Typescript> = ({
     diagnostics,
   } = ts.transpileModule(content, {
     fileName:
-      filename.slice(0, 1) === '.'
-        ? resolve(join(process.cwd(), filename))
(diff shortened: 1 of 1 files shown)


kaisermann (approved):

ashpieboop:
In my setup, some files are passed with relative path, no leading `./` (example: `build/views/home_dep.svelte`) so I believe your fix is incomplete.

src/transformers/typescript.ts
```diff
@@ -144,7 +144,10 @@ const transformer: Transformer<Options.Typescript> = ({
     sourceMapText: map,
     diagnostics,
   } = ts.transpileModule(content, {
-    fileName: filename,
+    fileName:
+      filename.slice(0, 1) === '.'
```
ashpieboop, line 148: replace with `!isAbsolute(filename)` (from `path`)

ashpieboop (changes requested):

src/transformers/typescript.ts
```diff
@@ -144,7 +144,10 @@ const transformer: Transformer<Options.Typescript> = ({
...
+        ? resolve(join(process.cwd(), filename))
```
ashpieboop, line 149: replace with `? resolve(join(basepath, filename))`

ashpieboop (changes requested):

src/transformers/typescript.ts
btakita: Applied in https://github.com/sveltejs/svelte-preprocess/commit/842867034f90c12d6a5fc523477133c8117b5d38. Thank you!

src/transformers/typescript.ts
btakita: Ditto. https://github.com/sveltejs/svelte-preprocess/commit/842867034f90c12d6a5fc523477133c8117b5d38

kaisermann:
@btakita Thanks for the contribution! However, I think that the last rebase/merge done added quite a few of undesired commits to the commit history of this branch 🤔

btakita:
@kaisermann np, I will remove them.

kaisermann:
@btakita There were still some extra commits there. I took the liberty of rewriting the branch's commit history. Could you check if the code is still as it should? If yes, we're good to go 👍

btakita:
@kaisermann lgtm. Thank you for removing those commits.

ashpieboop (approved):

kaisermann:
Released in 4.7.3 🎉

Issue #318, opened 2021-02-21
TypeScript: Better preprocessing to make importsNotUsedAsValues unnecessary


dummdidumm:
**Is your feature request related to a problem? Please describe.**
Currently, TypeScript is processed on a per-file-basis. Because the script tag is only part of the truth, a transformer was written to re-add imports which to TypeScript look unused but are actually used in the template. This means that people need to be very careful about how to write their imports, which can be especially cumbersome if a type import is from the same file as a real import.

```svelte
<script lang="ts">
  impor type { Type } from './foo';
  impor { Bar } from './foo';
</script>
```

Since TypeScript itself does not do that strict autocompletion separation (and they don't plan to change it, see [this issue](https://github.com/microsoft/TypeScript/issues/5858)), people have to do it by hand.

So what would be ideal is to 
- either somehow tell TypeScript which values are used so it does not throw them out, which would make the transformer unnecessary
- or write a more sophisticated transformer which traverses the AST and throws out all types from the imports (similar to #164, only without relying on a program if possible)

**Describe the solution you'd like**

Ideally, I don't have to be that strict with seperating type and value imports. I currently see two ways to get there.

### Solution 1: The Svelte-AST-compile-route

This solution is similar to the one currently used in `eslint-plugin-svelte3`. Some logic of #155 could maybe be reused.

1. get the whole Svelte file, blank style content
2. transpile script tags' TS to JS using the current transform which preserves all imports
3. do a svelte.compile which will return a list of variables with the info where they are used (module, template, etc)
4. append all variables to the bottom of the TS code so TS sees this "fake usage" and does not throw out imports that are only used in the template
5. transpile TS to JS without the "keep all imports" transformer

Advantages:
- probably the most correct and safest varaint

Disadvantages:
- Now TS needs the whole markup and would technically be a markup, not a script transformer. That's likely a breaking change
- It wouldn't work if people use markup that needs preprocessing, or that would need to happen before the script transformation

### Solution 2: Walk TS AST to find type only references

This solution would be similar to #164 . Basically, walk the AST and find all references of all imports, and filter those out who are type-only. Walking the AST without advanced capabilities like a type checker / TS program is likely tedious, but probably more robust. 

Advantages:
- no breaking change, only needs script contents

Disadvantages:
- maybe not possible without expensive type checker (although solution 1 would also be slower)
[...]

PR #342, opened 2021-04-14
WIP: Improve TypeScript imports preprocessing (#318) 

SomaticIT:
**/!\ This is a work-in-progress. DO NOT MERGE!**

### Introduction

I open this PR to init the work and the collaboration on a resolution for #318.

***Note:** This implementation requires an unreleased version of svelte to work.*
***Note 2:** You will find below a way to setup a test environment to test this implementation on your projects.*

### Background

Thanks to @dummdidumm advices, I added 3 features on `svelte`, some have been merged, some are waiting for review:
- [X] sveltejs/svelte#6169
- [ ] sveltejs/svelte#6192
- [ ] sveltejs/svelte#6194

These features enable 2 required things:
1. It makes `svelte` pass the full source markup to preprocessors
2. It allows to use `svelte.compile` to get a full report of used variables in markup

### Description

This PR makes two improvements.

First, it takes the markup passed by `svelte` and make it available to transformers.
This markup is optional here to be backward compatible with previous versions of `svelte`.

Then, it uses the full markup, strips the `script` and `style` tags and pass it to `svelte.compile` to get a report of all used variables in the template. This list of variables is appended to the TypeScript code to allow the TypeScript compiler to correctly handle imports. Thanks to this, we can remove the custom `importTransformer` when `markup` is available.

### TODO

This is a work in progress.
First of all, it needs above PRs to be merged and released on `svelte` side.

Then, it misses some cases:
- [x] Handling `context="module"` scripts
- [x] Stripping the injected code from sourcemaps
- [x] Adding more tests
- [ ] Testing on real-world project

For this last task, I will need the help of the community 😉.

I also have very few experience on sourcemap manipulation so if someone can help on this...

### Setup test environment

If you want to test working on this PR, you will need to do the following:
```bash
# Clone the fork with all svelte PRs merged
git clone https://github.com/SomaticIT/svelte.git
cd svelte

# Switch to the appropriate branch
git checkout @release/custom

# Prepare the project
npm install
npm run build

# Link the project globally
npm link

# Clone the fork of svelte-preprocess
cd ..
git clone https://github.com/SomaticIT/svelte-preprocess.git
cd svelte-preprocess

# Switch to the appropriate branch
git checkout @feature/typescript-imports

# Link svelte locally
npm link svelte
```

If you want to test this feature in your project (I will need your feedbacks):
```bash
# Do the previous steps then from the svelte-preprocess dir
npm run build
npm link

# In your project dir
npm link svelte
npm link svelte-preprocess
```

### Performances impact

On my computer, I benchmarked the preprocessing of this svelte file: `test/fixtures/TypeScriptImports.svelte`.
Before this PR: **~30ms**
After this PR: **~55ms**

- [X] It's really useful if your PR relates to an outstanding issue, so please reference it in your PR, or create an explanatory one for discussion. In many cases features are absent for a reason.
- [X] This message body should clearly illustrate what problems it solves. If there are related issues, remember to reference them.
- [X] Ideally, include a test that fails without this PR but passes with it. PRs will only be merged once they pass CI. (Remember to `npm run lint`!)

### Tests

- [X] Run the tests with `npm test` or `yarn test`

### DIFF
--- a/src/transformers/typescript.ts
+++ b/src/transformers/typescript.ts
@@ -2,16 +2,24 @@ import { dirname, isAbsolute, join } from 'path';
 
 import ts from 'typescript';
 import { compile } from 'svelte/compiler';
+import MagicString from 'magic-string';
+import sorcery from 'sorcery';
 
 import { throwTypescriptError } from '../modules/errors';
 import { createTagRegex, parseAttributes, stripTags } from '../modules/markup';
 import type { Transformer, Options } from '../types';
 
-type CompilerOptions = Options.Typescript['compilerOptions'];
+type CompilerOptions = ts.CompilerOptions;
+
+type SourceMapChain = {
+  content: Record<string, string>;
+  sourcemaps: Record<string, object>;
+};
 
 function createFormatDiagnosticsHost(cwd: string): ts.FormatDiagnosticsHost {
   return {
-    getCanonicalFileName: (fileName: string) => fileName,
+    getCanonicalFileName: (fileName: string) =>
+      fileName.replace('.injected.ts', ''),
     getCurrentDirectory: () => cwd,
     getNewLine: () => ts.sys.newLine,
   };
@@ -70,16 +78,37 @@ function getComponentScriptContent(markup: string): string {
   return '';
 }
 
+function createSourceMapChain({
+  filename,
+  content,
+  compilerOptions,
+}: {
+  filename: string;
+  content: string;
+  compilerOptions: CompilerOptions;
+}): SourceMapChain | undefined {
+  if (compilerOptions.sourceMap) {
+    return {
+      content: {
+        [filename]: content,
+      },
+      sourcemaps: {},
+    };
+  }
+}
+
 function injectVarsToCode({
   content,
   markup,
   filename,
   attributes,
+  sourceMapChain,
 }: {
   content: string;
   markup?: string;
-  filename?: string;
+  filename: string;
   attributes?: Record<string, any>;
+  sourceMapChain?: SourceMapChain;
 }): string {
   if (!markup) return content;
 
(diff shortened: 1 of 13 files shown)


src/transformers/typescript.ts
```diff
@@ -53,6 +54,45 @@ const importTransformer: ts.TransformerFactory<ts.SourceFile> = (context) => {
...
+  markup,
+  filename,
+}: {
+  content: string;
+  markup?: string;
+  filename?: string;
+}): string {
+  if (!markup) return content;
+
+  const { vars } = compile(stripTags(markup), {
```
dummdidumm, line 74: Since the Svelte compiler will throw errors on unknown properties, this needs to be surrounded with a try-catch for backwards-compatibility.

dummdidumm:
Thanks for taking a stab at this! 
Regarding `script context="module"`: In that case the whole instance script needs to be appended, too. To keep it consistent I suggest to separate the generated code with some comment like `// __svelte-preprocess__ start generated code` and split on that.
Regarding source maps: I don't know much about them either, I only know that mapping lines are separated by `;` , which could be used to determine where to cut off the mappings. If `sourcesContent` is returned, we can override that with the original code. I'm not sure what else we need to adjust.

src/transformers/typescript.ts
SomaticIT: If I'm correct, it will depend on the release of `svelte`. If the 3 PRs are merged and released in the same time, there will be no problem since the `markup` will not be passed so it will not try to compile it.

Am I correct?

dummdidumm: Ah right, yes that is correct. Should be documented in the code then just to be sure for future readers.

SomaticIT: Maybe we should inform `svelte` maintainers that the 3 PRs should be released together.
Do you agree?

*If it's not possible, using a try-catch will solve the issue*

SomaticIT:
Thank you for your feedbacks.

> To keep it consistent I suggest to separate the generated code with some comment like // __svelte-preprocess__ start generated code and split on that.

I was considering using something like `const $$$$$$$$$ = null;` to be sure that typescript will not remove it.
I also thought that this could allow us to retrieve it from the source map but I'm not sure that's the right way.

> Regarding script context="module": In that case the whole instance script needs to be appended, too.

I agree, I will add this with an appropriate test.

> Regarding source maps:...

I need to make some research on this..

Do you know a svelte maintainer who could help?

dummdidumm:
@milahu could know more

SomaticIT:
Hi @dummdidumm,

I added a test for `context="module"` script tags.
This seems to work well.

I noticed that we use common functions with `modules/markup.ts` so I also did a little refactoring to reuse the logic.

SomaticIT:
After some studies and tests, I managed to handle the source map transformation.

### Process

Here is a little summary of the process:
1. Inject variables and produce a source map
2. Transpile typescript and produce a source map
3. Strip variables and produce a source map
4. Concatenate the source map chain

*Note: For backward compatibility, this behavior is disabled if no `markup` is passed to the transformer.*

*Note 2: This process is only applied if the `sourceMap` option is `true`.
If not, it does not bother with sourcemaps and keep the previous behavior.*

### Refactoring

This implementation introduced a lot of complexity in the transformer.
So I splitted the code in the transformer to make it more readable.

### Drawbacks

This implementation introduces two new dependencies:
- `magic-string` allows to manipulate string with source map support. 
- `sorcery` allows to concatenate a source map chain.

*Fun fact: these libraries are written by Rich Harris.*

### Can we reuse these dependencies?

`magic-string` could be reused to handle source map in the `replace` transformer.

Maybe `sorcery` could be reused to concatenate source maps when multiple transformers are used during `autoProcess` (eg: `replace` + `typescript` + `babel`).

dummdidumm:
Now that the needed PRs are merged (new Svelte release pending), we can continue working on this. Some thoughts
- Since the "pass markup to preprocessors" PR was released already, this means we need the try-catch-approach to catch the "Svelte version does not support these options yet" case. That try-catch could be at the very top of the transformer, so if _anything_ breaks during the new approach we fall back to the old approach
- I _think_ (didn't test) it's possible to have an import inside the instance script that is only used within the module script block. If that's possible, then we need to append the module script to the instance script, too. I think appending is fine - this will result in invalid code but in "transpile only" mode TypeScript will just go through the stuff one-by-one

I'm on vacation the next 3 weeks so I can't look into this in the near future.

dummdidumm:
@SomaticIT I'll try to pick this up. The proposed solution is very smart, from what I see it only needs some minor adjustments.
- Since the three PRs in Svelte core were not merged as one, we need that `try-catch` solution
- We either need to add some logic to only return the transpiled instance/module script content, _or_ we could make this a markup transformer instead. The latter would break the order, but would make the code probably a little easier and faster because we don't need to do the work twice. @kaisermann thoughts?

SomaticIT:
@dummdidumm, thank you.

I'm currently on vacation. I will be back next week. I would be happy to help.

- I agree with the try-catch solution.

- I'm not sure I understand your second thought, if I remember correctly, only the transpiled instance/module script is already returned.

I also had an idea that could lead to better performance : maybe we could make svelte pass the `dev` option to preprocessor. This special transpilation process is only needed in dev mode since rollup drops unused artifacts. It will also avoid having a try-catch.

What do you think?

dummdidumm:
Ah yes you're right, the injected code is always below the real code, and it's stripped away. So this should work then.

I tested this out on a project and ran into one issue: When there's no code after transpilation, because the code is type-defs only, the mappings are an empty string, which breaks `sorcery` (also see https://github.com/Rich-Harris/sorcery/issues/167). A check is needed to return early in this case.
Another thing is: When using the TypeScript transpiler, it's now necessary to also install `source-map` and `sorcery`, which is somewhat of a breaking change. @kaisermann what's the best strategy here? Should we add these libraries as `dependencies`, or bundle them, or request the user to install them themselves?

src/transformers/typescript.ts
```diff
@@ -167,6 +293,110 @@ const transformer: Transformer<Options.Typescript> = ({
...
+    sourceMapChain,
+  });
+
+  const code = stripInjectedCode({
+    transpiledCode,
+    markup,
+    filename,
+    sourceMapChain,
+  });
+
```
dummdidumm, line 393: This fixes the sorcery error that happens on empty code/maps. A test case should be added to handle this.

```suggestion

if (!code) {
  return { code, diagnostics };
}

```

kaisermann:
I've been a bit AFK these last two months, so I'll need to get my bearings regarding this whole scenario again.

> it's now necessary to also install source-map and sorcery, which is somewhat of a breaking change. @kaisermann what's the best strategy here?

We are already "educating" people to install what they're going to need for their respective workflows. However, I'm also all up for simplicity and, as it seems, most people are using/will use Svelte with Typescript at some point. I'd like to keep only the "main tools" as "peer dependencies" (ts, postcss, stylus, sass, etc) if possible, although `postcss-load-config` kind of already goes against this 🤔 .

What's keeping us from having the `advanced type import` as the default behaviour? Just these dependencies or does it behaves differently in other ways? Is there an advantage for having both the current and new behaviour? 

> However, if TypeScript implements support for strict separation of types and values (PR pending, maybe it's part of 4.5)

Can you link the PR here?

dummdidumm:
@kaisermann this can be closed now

PR #370, opened 2021-06-06
docs: adjust config for esm usage

### LINKED ISSUE #274: Reccomend svelte.config.js
Currently the docs only mention `svelte.config.js` as an integration step for VS-Code. 
It is not even mentioned in the readme, despite the fact it contains enough core to set up the build step for those already familure with the preprocessors they wish to use (everybody who wants to use a preprocessor, excluding first timers to a given language).
This makes it super easy for new users to miss the docs and is a semi-frequent cause of confused users in discord.

I believe it would be a prudent to *always* suggest it as the place to configure things.
It is recommended by language-tools, necessary for the only officially supported editor/plugin,
and is currently the best-practice for easy integration/switching of build tools (snowpack even reads it on its own).
If/when we finally get an official preprocessor cli, the file will also be wanted/required by it.

The svelte ecosystem revolves around this file already, so treating it as *the standard*
instead of an extra step seems like the way to go.
For conditional configuration, the `createPreprocessors` function suggested by language tools could be reused.
If you want to ensure users know that the file isn't required, a small note that you can also do this directly in rollup/webpack/snowpack could be added in small letters at the bottom of some deeply hidden doc page.

* Directly link the related language-tools docs.
* Leave it as is and hope users eventually figure it out on their own.

**How important is this feature?**
Moderate. This affects most people trying to use preprocessors and there are probably several who followed the readme, got build working, and simply live with the broken ide/editor experience believing its NYI.

### LINKED ISSUE #367: Use with sveltekit
It's a problem for me, but maybe the answer is No?

Documentation on usage with sveltekit, which doesn't use rollup.

Very important. sveltekit is the future of svelte.

ignatiusmb:
Closes #367 
Closes #274

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,17 +38,16 @@ Writing your own preprocessor for, i.e SCSS is easy enough, but it can be cumber
 
 `svelte-preprocess` is a custom svelte preprocessor that acts as a facilitator to use other languages with Svelte, providing multiple features, sensible defaults and a less noisy development experience.
 
-```js
-import svelte from 'rollup-plugin-svelte'
-import sveltePreprocess from 'svelte-preprocess'
+It is recommended to use with `svelte.config.js` file, located at the project root. For other usage, please refer to [usage documentation](#usage-documentation).
 
-export default {
-  plugins: [
-    svelte({
-      preprocess: sveltePreprocess({ ... })
-    })
-  ]
+```js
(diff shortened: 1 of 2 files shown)


kaisermann (approved):

kaisermann:
Nice! Thanks 🎉

PR #380, opened 2021-06-30
Fixes prepend data being injected in other languages

### LINKED ISSUE #379: SCSS prependData should happen only when `lang="scss"` or `lang="sass"`
**Describe the bug**
Currently, when we have `prependData` option of `scss` preprocessor, the `lang` type of `<style>` is not being respected, the `prependData` are being injected ignoring the `lang` type. The `prependData` logic should be executed only when `lang="scss"` or `lang="sass"`.

**To Reproduce**
https://github.com/maykefreitas/svelte-preprocess-example

**Expected behavior**
The expected behavior would be to not inject `prependData` unless the `style` tag has `lang="scss"`.

- svelte-preprocess version: latest

The cause of this bug is the fact that [this logic](https://github.com/sveltejs/svelte-preprocess/blob/main/src/processors/scss.ts#L26) should be executed after the `if` declaration, not before.

maykefreitas:
Closes #379

### DIFF
--- a/src/processors/coffeescript.ts
+++ b/src/processors/coffeescript.ts
@@ -15,6 +15,10 @@ export default (options?: Options.Coffeescript): PreprocessorGroup => ({
       dependencies,
     } = await getTagInfo(svelteFile);
 
+    if (lang !== 'coffeescript') {
+      return { code: content };
+    }
+
     content = prepareContent({
       options: {
         ...options,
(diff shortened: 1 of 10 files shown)


kaisermann:
Hey, @maykefreitas 👋 

Thanks for the contribution! Can you also do this for the other processors such as `stylus`, `less` and `coffeescript`?

maykefreitas:
hey @kaisermann, sure I can :)

maykefreitas:
@kaisermann done ✅

kaisermann (approved):
Neat 👌

PR #385, opened 2021-07-15
fix: ts compilerOptions cannot be generated correctly

pikadun:
`options.tsconfigFile` is optional and has no default value.
When tsconfigFile is undefiend, compilerOptions cannot be generated correctly.

- [ ] This message body should clearly illustrate what problems it solves. If there are related issues, remember to reference them.

### Tests


dummdidumm:
I think this fix is unnecessary. The equality check checks for strict equality, so as long as `tsconfigFile` is not set to `false`, it will go into that branch. The docs don't mention that you can set this to `false`, but according to the typings you can.

kaisermann:
Yeah, IIRC allowing `false` was to allow one to disable the config file lookup.

PR #392, opened 2021-08-22
Better Typescript transpilation

### LINKED ISSUE #318: TypeScript: Better preprocessing to make importsNotUsedAsValues unnecessary

dummdidumm:

### Tests

This PR builds upon the stellar work by @somaticIt in #318 . See that PR for more info on the "why do this"?

This PR superseeds #318 and adds the following additions:
- moves new functionality behind a flag. Dynamically imports required new libraries. This makes the change backwards compatible
- adds a check and test for the case where TS code consists of types only and has no content after transpilation
- merge main

Closes #318


dummdidumm:
The failing tests seem unrelated, I guess they appear as a cause of the Svelte dev dependency bump.
I also noticed that `build` will no longer work without `"skipLibCheck": true` in the `tsconfig.json` because of some type errors in the latest Svelte version.

src/types/options.ts
```diff
@@ -74,6 +74,7 @@ export type Typescript = {
   tsconfigFile?: string | boolean;
   tsconfigDirectory?: string | boolean;
   reportDiagnostics?: boolean;
+  useAdvancedImportTranspiler?: boolean;
```
dummdidumm, line 77: Don't know if this is the best name for the option, better names welcome

SomaticIT: Do we really need a new option? This new process should impact existing codebases.

kaisermann:
Neat! I've been a  bit AFK these last two months, so I'll need to get my bearings regarding this whole scenario again, but it looks promising!

In time: what's the relation between #342 and this PR? Can we close #342?

src/types/options.ts
dummdidumm: My arguments in favor of the option:
- backwards compatible. If this is the new default, then either it's a breaking change because people would need to install sorcery and magic-string, or we add them to `dependencies` but then everyone, even those who don't use TypeScript, need to download the additional packages
- intermediate solution: there are plans from the TypeScript team to provide a new import option which would strictly separate types and values, which would make this advanced transpilation unnecessary. If that solution is implemented, people are free to choose the solution right here or use the new TypeScript option.

dummdidumm:
> In time: what's the relation between #342 and this PR? Can we close #342?

Somewhat related. If the direction of adding a new option is accepted, then yes #342 can be closed. If we want to make this new transpilation the new default behavior, we first need to change the implementation and that could happen in #342 as well. For my arguments in favor of making this an option, see the comment above.

src/types/options.ts
kaisermann: FWIW adding both as dependencies adds +- 1.6mb to the node_modules (ignoring possible dep deduplication).

kaisermann: I like the idea of having an intermediate solution. If there's a possibility that this can be solved in a future ts version, we should plan this change with that in mind. If we go through with this, should we say anything about it being a temporary solution?

test/fixtures/TypeScriptImports.svelte
```diff
@@ -0,0 +1,109 @@
+<script lang="ts">
```
kaisermann, line 1: Is there a reason for this fixture to be this big?

SomaticIT: It was my decision, because I wanted to (manually) ensure that all features are taken into account and that the sourcemap is well handled.

SomaticIT: Maybe it could be reduced now...

src/types/options.ts
dummdidumm: Good idea about adding a note. I'm not 100% sure if/when the TS solution lands, so the note should reflect that. And even after that, people might prefer not strictly separating types and values and still want to choose this "temporary" solution.

SomaticIT:
Hi @dummdidumm,

Thank you for taking some time on this PR. Your evolution is great.

Here is my feedback:

1. I think that both `sorcery` and `magic-string` could be added as dependencies (or bundled). They are very small and could be used by others transformers (like `replace`). Moreover, I think that forcing people to install dependencies manually is a source of errors.

2. I'm not sure that we need to introduce a new option. During my tests, I tried this new preprocessor on a big project and it doesn't create any breaking changes. Maybe we could take some time to investigate on more projects to be sure but I think it's safe.

3. We could improve the preprocessor performance by using the new *advanced* transpiler only when running in `dev` mode. When compiling for production, `rollup` will automatically remove unused imports so this will generates a runtime issue. **It forces us to create a new PR on `svelte` to pass the `dev` option to preprocessors.**


So I have 3 questions:
1. Could we include `sorcery` and `magic-string` as dependencies (or bundle them)?
2. Could we safely remove the option?
3. Could we improve the performance by running this transpiler only in `dev` mode?

dummdidumm:
>     1. I think that both `sorcery` and `magic-string` could be added as dependencies (or bundled). They seems to be small and could be used by others transformers (like `replace`). Moreover, I think that forcing people to install dependencies manually is a source of errors.

They already have to for `typescript`, so the mode of adding new dependencies should be familiar.

>     2. I'm not sure that we need to introduce a new option. During my tests, I tried this new preprocessor on a big project and it doesn't create any breaking changes. Maybe we could take some time to investigate on more projects to be sure but I think it's safe.

I trief this, too, and didn't find any breaking things, so I'm also very confident that this works. The argument for making it an option is not related to this for me.

>     3. We could improve the preprocessor performance by using the new _advanced_ transpiler only when running in `dev` mode. When compiling for production, `rollup` will automatically remove unused imports so this will generates a runtime issue. **It forces us to create a new PR on `svelte` to pass the `dev` option to preprocessors.**

I think we shouldn't make these assumptions. Preprocessors could run in other contexts unrelated to Rollup or without removing unused imports as well.



My arguments in favor of the option:

-    backwards compatible. If this is the new default, then either it's a breaking change because people would need to install sorcery and magic-string, or we add them to dependencies but then everyone, even those who don't use TypeScript, need to download the additional packages
-    this is somewhat of an intermediate solution: there are plans from the TypeScript team to provide a new import option which would strictly separate types and values, which would make this advanced transpilation unnecessary. If that solution is implemented, people are free to choose the solution right here or use the new TypeScript option.

SomaticIT:
> They already have to for typescript, so the mode of adding new dependencies should be familiar.

You're right, but it means that you now need to install 3 additionnal dependencies (typescript, magic-string, sorcery). Installing typescript is necessary when you use typescript in a project so it's normal to install it manually. However installing magic-string and sorcery is more obscure.

> backwards compatible. If this is the new default, then either it's a breaking change because people would need to install sorcery and magic-string, or we add them to dependencies but then everyone, even those who don't use TypeScript, need to download the additional packages

I'm in favor of adding them to dependencies. In [this comment](https://github.com/sveltejs/svelte-preprocess/pull/342#issuecomment-820741262), I explain that both dependencies can be reused to improve the sourcemap handling of `svelte-preprocess`. (magic-string for replace and sorcery to combine multiple transformers sourcemaps).

> this is somewhat of an intermediate solution: there are plans from the TypeScript team to provide a new import option which would strictly separate types and values, which would make this advanced transpilation unnecessary. If that solution is implemented, people are free to choose the solution right here or use the new TypeScript option.

I'm not totally sure that this will be an intermediate solution. Even if TypeScript provides a new import option to strictly separate types and values, this will generate additional code and will make imports less readable. Those who are using TypeScript for long time are used to combine types and values in the same import statement and IMHO, they will prefer staying like this.

dummdidumm:
> > backwards compatible. If this is the new default, then either it's a breaking change because people would need to install sorcery and magic-string, or we add them to dependencies but then everyone, even those who don't use TypeScript, need to download the additional packages
> 
> I'm in favor of adding them to dependencies. In [this comment](https://github.com/sveltejs/svelte-preprocess/pull/342#issuecomment-820741262), I explain that both dependencies can be reused to improve the sourcemap handling of `svelte-preprocess`. (magic-string for replace and sorcery to combine multiple transformers sourcemaps).

In the case of `magic-string` that's certainly true, it could be used in other scenarios. `sorcery` - not so much, as combining multiple sourcemaps is already handled by Svelte core. That said, the functionality from Svelte core could _maybe_ be exported to be used.
 
> > this is somewhat of an intermediate solution: there are plans from the TypeScript team to provide a new import option which would strictly separate types and values, which would make this advanced transpilation unnecessary. If that solution is implemented, people are free to choose the solution right here or use the new TypeScript option.
> 
> I'm not totally sure that this will be an intermediate solution. Even if TypeScript provides a new import option to strictly separate types and values, this will generate additional code and will make imports less readable. Those who are using TypeScript for long time are used to combine types and values in the same import statement and IMHO, they will prefer staying like this.

I get that sentiment. We should rename the option to something that doesn't sound like it's superior to the current solution, like `handleMixedImports`. I still vote that it's an option. If we make it the default, then there needs to be an additional check for the minimum Svelte version installed in the repository, or everything is surrounded with `try-catch`, which could get expensive performance-wise.

@kaisermann you asked about the TypeScript PR: https://github.com/microsoft/TypeScript/pull/44619

kaisermann:
> However installing magic-string and sorcery is more obscure.

Agreed, seems too specific and obscure. Also, I've been wanting to use `magic-string` in the `replace` transformer for a while now, so adding it as a dependency feels natural. 


> We should rename the option to something that doesn't sound like it's superior to the current solution, like `handleMixedImports`

Maybe `allowMixedImports`? Anyway, I like this naming direction a bit more than `useAdvanced...`, which seems too generic (advanced how?). 

> If we make it the default, then there needs to be an additional check for the minimum Svelte version installed in the repository

This wouldn't be the first time we did this: https://github.com/sveltejs/svelte-preprocess/blob/6346194254/src/autoProcess.js#L92. 

We can have it now as an opt-in feature and wait for a bit to see how it fairs in the wild and set it as the default behaviour on a future minor or possibly major version.

dummdidumm:
I changed the option name to `handleMixedImports` (`allowMixedImports` is just as good for me, you decide) and made `sorcery` and `magic-string` "real" dependencies.
Last question is whether or not to make this the default. Since the don't have the dependency problem anymore and since we can check for the Svelte version, I'm slightly in favor of making it the new default for everyone who has the appropriate minimum Svelte version.

SomaticIT:
It seems good for me.
I will give it a try on a large project and give you my feedback.
Hope I have time tomorrow.

src/transformers/typescript.ts
```diff
@@ -53,70 +69,161 @@ const importTransformer: ts.TransformerFactory<ts.SourceFile> = (context) => {
...
+    generate: false,
+    varsReport: 'full',
+    errorMode: 'warn',
+    filename,
+  });
 
-  basePath = dirname(tsconfigFile);
+  const sep = '\nconst $$$$$$$$ = null;\n';
+  const varsValues = vars.map((v) => v.name).join(',');
+  const injectedVars = `const $$vars$$ = [${varsValues}];`;
```
dummdidumm, line 130: Who knows, maybe some unknown globals/values are used in the markup, and we don't want to add unnecessary diagnostics
```suggestion
  const injectedVars = `// @ts-ignore
const $$vars$$ = [${varsValues}];`;
```

SomaticIT: @dummdidumm, this is not necessary. Since we are only transpiling the code, there is no type checking...

dummdidumm: You are right, the only diagnostics that are returned are syntax errors which do not happen in that area. Undone.

dummdidumm:
@SomaticIT were you able to have a look? Would love to get this merged. @kaisermann any change request for the PR or is this good to go for you?

kaisermann:
I will give the transformer a final look, format the file and we should be good. I should be able to do this in the following days 👍

SomaticIT:
@dummdidumm, sorry for the delay, I had a lot of work these past days.

I took some time this morning to try it out.
It works well on many modules but we have an issue with imported `stores`.

---

### Reproduction (maybe):

```svelte
<script lang="ts">
  import { store } from "somelib";
</script>

Store value: {$store}
```

This will throw: `Uncaught ReferenceError: store is not defined`.

---

I think this is because the `store` variable is referenced as `$store` in the transpiled code so the import is stripped by TypeScript.
I will investigate more to find a solution and let you know.

dummdidumm:
No worries about the delay!

That's a good catch, I can reproduce this with a test. I'll fix this by searching all referenced variabled for a `$`-prefix and append the variables without the prefix.

SomaticIT:
I updated my local project with your changes. It works fine with *template-only* stores but not with *code-only* stores.
Also, the `test/fixtures/store.ts` file is missing.

I tried to run the updated `svelte-preprocess` on my project but I also have store-subscriptions inside some scripts.
I created a little patch to add the missing `store.ts` and add the test case with fails.

You can find it here: [typescript-test.patch](https://github.com/sveltejs/svelte-preprocess/files/7085550/typescript-test.zip)

SomaticIT:
OK, I progress!!

I found a (tricky) way to extract *store-subscriptions* from script.
It allowed me to finally compile my large project.

Here is the patch: [typescript-inject-code-stores.patch](https://github.com/sveltejs/svelte-preprocess/files/7085754/typescript-inject-code-stores.zip)

If you apply it after the previous one, the failing test will also resolve.
In this patch, I also improved the way you extracted store-subscriptions from `varnames` by doing it directly in the first loop.

I still have an issue with the code.
On my project, all comments are now replaced by a text `undefined` in the HTML:

---

### Example

```svelte
<script lang="ts">
  let hello = "world";
</script>

<p>{hello}</p>
<!--test comment-->
<p>{hello}</p>
```

will produce 

```
world
undefined
world
```

I'm not sure this is a direct impact of this PR. I will try to investigate.

SomaticIT:
Victory!!!

I finally found what was causing the comment issue, it was my fault :(.
Indeed, when I extracted the tag regex from `transformMarkup` to `createTagRegex` in `src/modules/markup.ts`, I forgot the first `/`.

This was making markup transform to handle comments as `<template>` tag.

@dummdidumm, you can find the fix in the last commit of the PR I submitted on your fork.

dummdidumm:
Thank you for looking into this! I found one more edge case I fixed: It's valid to import something in the instance script but only use it in the module script. Therefore I append the module script below the instance script now. @kaisermann this is now good to go (the lintings should be fine now, too), please make sure that the "co-authored by"-message for @SomaticIT is part of the squash-merge-commit so he gets proper credit.

test/fixtures/TypeScriptImports.svelte
SomaticIT: Finally, I think that we should keep this fixture intact, it is very useful to reproduce edge cases.

SomaticIT:
I updated my local version and everything is working fine!
Congratulations for your work!

I was happy to collaborate with you on this very exciting feature!!

SomaticIT:
@dummdidumm, I found one more edge case: the codestores detection regex can detect reserved keywords.
I fixed it by filtering reserved keywords from the result.

#### Patch

You can find the patch here: [code-vars-filter-reserved-keywords.patch](https://github.com/sveltejs/svelte-preprocess/files/7105092/code-vars-filter-reserved-keywords.zip)

#### Exemple:

```typescript
const filter = { $in: ["test"] };
```

In the above code, the codestores regex will detect `$in` as a potential store so it will append `in` to the injected vars but `in` is a reserved keyword in javascript. It is safe to remove it since it's not possible to create a store with a reserved keyword.

src/transformers/typescript.ts
```diff
@@ -53,70 +69,175 @@ const importTransformer: ts.TransformerFactory<ts.SourceFile> = (context) => {
...
+  );
 
-  if (error) {
-    throw new Error(formatDiagnostics(error, basePath));
+  const varsString = [...codestores, ...varnames].join(',');
+  const injectedVars = `const $$vars$$ = [${varsString}];`;
+  const injectedCode =
+    attributes?.context === 'module'
+      ? `${sep}${getScriptContent(markup, false)}\n${injectedVars}`
+      : `${sep}${getScriptContent(markup, true)}\n${injectedVars}`;
```
kaisermann, line 147:
```suggestion
const injectedCode = `${sep}${getScriptContent(
  markup,
  attributes?.context === "module"
)}\n${injectedVars}`;
```

src/transformers/typescript.ts
```diff
@@ -175,11 +312,192 @@ const transformer: Transformer<Options.Typescript> = ({
...
+  return {
+    code: transpiledCode,
+    map: sourceMapText,
+    diagnostics,
+  };
+}
+
+const transformer: Transformer<Options.Typescript> = async ({
+  content,
+  filename = 'source.svelte',
```
kaisermann, line 459: Why the default value? For less verbose tests?

src/transformers/typescript.ts
```diff
@@ -175,11 +312,192 @@ const transformer: Transformer<Options.Typescript> = ({
...
+  markup,
+  options = {},
+  attributes,
+}) => {
+  const basePath = process.cwd();
+
+  filename = isAbsolute(filename) ? filename : resolve(basePath, filename);
+
+  const compilerOptions = getCompilerOptions({ filename, options, basePath });
+  const canUseMixedImportsTranspiler = +pkg.version.split('.')[1] >= 39;
```
kaisermann, line 469: Would it make sense to check for the major segment as well? Conceptually, just checking for the minor segment seems weird, albeit it's probably good enough 🤔

kaisermann (approved):
Just added some minor comments. Other than that this looks great, good job @SomaticIT and @dummdidumm 🎉

src/transformers/typescript.ts
dummdidumm: @somaticit added this, but I don't know why TBH 😄 I removed it and the tests pass.

SomaticIT: If I remember correctly, it was to help running the transformer in standalone mode for my local tests.
I think it's safe to remove

SomaticIT:
It seems good for me, I'll continue to use this specific version on my largest project to see if I discover more edge cases. However it seems to work well for me.

kaisermann (approved):

kaisermann:
Released as v4.9.0 🎉  Awesome job 🙏

janosh:
> - moves new functionality behind a flag

Are you referring to `"useAdvancedImportTranspiler": true`? Where does this go? Putting it in `tsconfig.json` gives me

> Unknown compiler option 'useAdvancedImportTranspiler'

on latest `typescript@4.4.3`.

dummdidumm:
It's called `handleMixedImports`. `sveltePreprocess({typescript: {handleMixedImports: true}})` in your `svelte.config.js` . Note that it's automatically enabled if you have at least Svelte 3.39

janosh:
@dummdidumm Looks like this change does not apply to imports from `'@sveltejs/kit'` inside `.svelte` files.

```
// src/routes/index.svelte
import type { LoadInput, LoadOutput } from '@sveltejs/kit' // works
import { LoadInput, LoadOutput } from '@sveltejs/kit' // errors
```

> Failed to resolve entry for package "@sveltejs/kit". The package may have incorrect main/module/exports specified in its package.json: Missing "." export in "@sveltejs/kit" package


Is this expected?

See https://github.com/sveltejs/kit/issues/2384#issuecomment-933316317 for details.

dummdidumm:
This is expected. If your import consists of types only, you need to add use `import type` . If there's at least one value import, you can now mix them using `import` , which wasn't possible previously.

janosh:
I see. Prob should have read this thread more carefully.

PR #406, opened 2021-09-10
(fix) Clear outDir if set in tsconfig

### LINKED ISSUE #405: Typescript processing breaks when tsconfig has `outDir` set
**Describe the bug**

When `outDir` is set in tsconfig.json, Typescript generates a `sourceMapChain` with a bad path for the `sources` in the source map, which leads to an ENOENT error somewhere in `sorcery` when it tries to open the file.

**Logs**

I finally traced it down to this `sourceMapChain` (extraneous bits omitted)

```js
{
  filename: '/home/dimfeld/cv/code/apps/data-admin-tool/src/routes/customer_usage/index.svelte',
  sourcemaps: {
     '/home/dimfeld/cv/code/apps/data-admin-tool/src/routes/customer_usage/index.svelte.injected.ts': SourceMap {
   version: 3,
   file: 'index.svelte.injected.ts',
   sources: [ 'index.svelte' ],
   sourcesContent: [ null ],
   },
    '/home/dimfeld/cv/code/apps/data-admin-tool/src/routes/customer_usage/index.svelte.transpiled.js': {
   version: 3,
   file: 'index.svelte.injected.js',
   sourceRoot: '',
   // Here's the bad value
   sources: [ '../src/routes/customer_usage/index.svelte.injected.ts' ],
}
```

which leads to this error:
```
Error: ENOENT: no such file or directory, open '/home/dimfeld/cv/code/apps/data-admin-tool/src/routes/src/routes/customer_usage/index.svelte.injected.ts'
```

Note the double `src/routes` that comes from resolving the `sources` value. When `outDir` is not set, the sources value is just `['index.svelte.injected.ts']`.

**To Reproduce**

Build https://github.com/dimfeld/svelte-preprocess-injected-bug

**Expected behavior**

Doesn't crash.

- svelte-preprocess 4.9.4, reproduces on >= 4.9.0. I think the underlying issue was there before but the older source map processing didn't actually throw an exception.
- svelte 3.42.4
- Ubuntu 21.04
- Whether your project uses Webpack or Rollup -- Reproduces on both.

**Additional context**

In my case this happens in a monorepo when loading Svelte components from another package that also contains regular Typescript files, so outDir is set in a package for compiling those files. Most normal Svelte projects wouldn't be using outDir so this probably doesn't come up much.

I think the best course of action here is for the typescript transformer to alter the loaded Typescript configuration to remove outDir unconditionally. Happy to submit a PR if you agree.

dimfeld:
As mentioned in #405. Since it's a quick fix I decided to just submit the PR before any discussion. Happy to discuss more if you're not sure if this is the right thing to do.

If outDir is set, Typescript sets directories in the source maps relative to it
which causes problems in `concatSourceMaps` when the paths in the source
point to nonexistent files instead of the virtual modules set up by the
transformer.

Tested against https://github.com/dimfeld/svelte-preprocess-injected-bug and also added a new unit test.

Fixes #405

### DIFF
--- a/src/transformers/typescript.ts
+++ b/src/transformers/typescript.ts
@@ -253,6 +253,8 @@ function getCompilerOptions({
   const compilerOptionsJSON = {
     moduleResolution: 'node',
     target: 'es6',
+    // Clear outDir since it causes source map issues when the files aren't actually written to disk.
+    outDir: undefined,
   };
 
   Object.assign(compilerOptionsJSON, options.compilerOptions);
--- a/src/transformers/typescript.ts
+++ b/src/transformers/typescript.ts
@@ -253,8 +253,6 @@ function getCompilerOptions({
   const compilerOptionsJSON = {
     moduleResolution: 'node',
     target: 'es6',
-    // Clear outDir since it causes source map issues when the files aren't actually written to disk.
-    outDir: undefined,
   };
 
   Object.assign(compilerOptionsJSON, options.compilerOptions);
@@ -272,6 +270,8 @@ function getCompilerOptions({
     ...(convertedCompilerOptions as CompilerOptions),
     importsNotUsedAsValues: ts.ImportsNotUsedAsValues.Error,
     allowNonTsExtensions: true,
+    // Clear outDir since it causes source map issues when the files aren't actually written to disk.
+    outDir: undefined,
   };
 
   if (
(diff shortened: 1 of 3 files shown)


dummdidumm (approved):
As discussed in the linked issue, this is a good solution for me.

kaisermann:
Thanks 🎉  Released in v4.9.5

PR #407, opened 2021-09-12
(fix) return default settings when no tsconfig found

dummdidumm:
This prevents the code from being transpiled as es3 in this situation, which makes the `const $$$$$ = ..` separator get transformed into `var $$$$$ = ...` which then is no longer split out. This situation came up in the tests of `svelte-jester`.
No test added, because it's not really possible, because there's a tsconfig in this repo which would be found.

### Tests


benmccann (approved):

kaisermann (approved):

kaisermann:
Thanks 🎉  Released in v4.9.5

PR #408, opened 2021-09-15
Don't overwrite `target` settings from tsconfig.json

### LINKED ISSUE #383: TypeScript processor reads and parses tsconfig for each processed file; target is not respected
**Describe the bug**
For performance reasons, tsconfig should be read only once per project. This can be verified in the debugger.
Also, target from tsconfig should be respected. Currently it is hardcoded to 'es6' no matter what tsconfig specifies.

- `svelte-preprocess` version 4.7.3
- Whether your project uses Webpack or Rollup: Snowpack

**Workaround for setting target in config:**

```preprocess({typescript: {compilerOptions: {target: 'esnext'}}})```

However, something still re-processes it again, at least when using Snowpack. Will investigate more.

### PR DESCRIPTION
**Pull request by dimfeld**
Resolves #383.

I moved all of the default options down to line 269, since the current behavior gave unusual precedence to the defaults where they would override options from tsconfig, but not options directly passed to the preprocessor, which feels unintuitive. I can move the `moduleResolution` setting back though, if you disagree.

~~This could be considered semver-major, since moving the default target to ES2021 would cause problems if anyone was inadvertently relying on ES6 transpilation to demodernize their JS and didn't have a proper Babel step or something in the bundler.~~

### DIFF
--- a/src/transformers/typescript.ts
+++ b/src/transformers/typescript.ts
@@ -249,24 +249,20 @@ function getCompilerOptions({
   options: Options.Typescript;
   basePath: string;
 }): CompilerOptions {
-  // default options
-  const compilerOptionsJSON = {
-    moduleResolution: 'node',
-    target: 'es6',
-  };
-
-  Object.assign(compilerOptionsJSON, options.compilerOptions);
+  const inputOptions = options.compilerOptions ?? {};
 
   const { errors, options: convertedCompilerOptions } =
     options.tsconfigFile !== false || options.tsconfigDirectory
-      ? loadTsconfig(compilerOptionsJSON, filename, options)
-      : ts.convertCompilerOptionsFromJson(compilerOptionsJSON, basePath);
+      ? loadTsconfig(inputOptions, filename, options)
+      : ts.convertCompilerOptionsFromJson(inputOptions, basePath);
 
   if (errors.length) {
     throw new Error(formatDiagnostics(errors, basePath));
   }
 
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: src/transformers/typescript.ts [hunk 1]
```diff
@@ -249,24 +249,20 @@ function getCompilerOptions({
...
-      : ts.convertCompilerOptionsFromJson(compilerOptionsJSON, basePath);
+      ? loadTsconfig(inputOptions, filename, options)
+      : ts.convertCompilerOptionsFromJson(inputOptions, basePath);
 
   if (errors.length) {
     throw new Error(formatDiagnostics(errors, basePath));
   }
 
   const compilerOptions: CompilerOptions = {
+    target: ts.ScriptTarget.ES2021,
```
dummdidumm, line 264: Strictly speaking this is a breaking change for everyone who did not define a target before. I would change this to be es6 like before. Everyone who sets it otherwise is then able to alter it (which isn't possible currently because of the forced option).

dummdidumm (changes requested):
Looks good to me apart from the (in my view) breaking change.

src/transformers/typescript.ts
dimfeld: Ah yes, of course that makes sense.

dimfeld:
Updated to set default target to ES6 if none is present anywhere else. Thanks!

dummdidumm (approved):

kaisermann (approved):
Nice! Thanks 🙏

PR #416, opened 2021-09-27
fix: coffeescript sourcemap

### LINKED ISSUE #413: get error "Cannot read properties of undefined (reading 'length')"  in sveltekit when use coffeescript
routes/index.svelte
```
<script lang="coffee">
text = 'hello'
</script>

<h1>{text}</h1>
```

svelte.config.js
```
import preprocess from 'svelte-preprocess'

/** @type {import('@sveltejs/kit').Config} */
const config = {
  kit: {
    // hydrate the <div id="svelte"> element in src/app.html
    target: '#svelte',
  },
  preprocess: preprocess({ coffeescript: true }),
}
export default config
```

I really like coffee, hope this would be fixed faster, thanks.

bluwy:
Fixes #413

The coffeescript transformer was using an old sourcemap format I believe. It was something like:

```
SourceMap {
  lines: [
    LineMap { line: 0, columns: [Array] },
    <1 empty item>,
    LineMap { line: 2, columns: [Array] }
  ]
}
```

This change uses the `v3SourceMap` string from the compile result instead.

### Tests

### DIFF
--- a/src/transformers/coffeescript.ts
+++ b/src/transformers/coffeescript.ts
@@ -17,10 +17,12 @@ const transformer: Transformer<Options.Coffeescript> = ({
   } as Omit<Options.Coffeescript, 'bare'>;
 
   if (coffeeOptions.sourceMap) {
-    const { js: code, sourceMap: map } = coffeescript.compile(
+    const { js: code, v3SourceMap } = coffeescript.compile(
       content,
       coffeeOptions,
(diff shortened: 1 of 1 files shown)


kaisermann (approved):

kaisermann:
Thanks! Released in `v4.9.6` 🎉

PR #417, opened 2021-09-30
fix: deprecate type attribute

### LINKED ISSUE #414: Deprecate type=".." and several lang=".."
Right now it's possible to signal that TS is used in 4 ways
- `lang="ts"`
- `lang="typescript"`
- `type="text/ts"`
- `type="text/typescript"`

The first of these is the recommended way, and Vite only accepts this version (see https://github.com/vitejs/vite/pull/5105 for more). It's confusing to have multiple ways to express the same and puts the burden on all other tools to check these four values.

Deprecate all `type=".."` usages. Deprecate `lang="typescript"` and `lang="javascript"`. Remove handling these in 5.0

Deprecate, but don't remove

Not important now, but important in the long run for standardization

kaisermann:
FIxes #414

- [x] Deprecate `lang="typescript"`
- [x] Deprecate `type="text/ts"`
- [x] Deprecate `type="text/typescript"`

I think the language detection and the auto preprocessor are still in the need of a significant refactor, but this should be enough for now.


dummdidumm (commented):
I would add a global to `src/modules/language.ts` that checks if the warning was already thrown, to only throw it once per npm process to prevent pollution of logs. Other than this this looks good to me, thank you very much 👍

dummdidumm (approved):

kaisermann:
Released in `v4.9.8` 🎉

PR #434, opened 2021-11-18
(feat) support preserveValueImports introduced in TS 4.5

dummdidumm:
This new flag keeps imports as is, essentially functioning the same as the import transformer.

### Tests


bluwy (approved):

test/transformers/typescript.test.ts
```diff
@@ -140,6 +140,23 @@ describe('transformer - typescript', () => {
...
+
+      const opts = sveltePreprocess({
+        typescript: {
+          tsconfigFile: false,
+          compilerOptions: { preserveValueImports: true },
+        },
+      });
+
+      const { code } = await preprocess(tpl, opts);
+      expect(code).toContain("import { Bar } from './somewhere'");
```
kaisermann, line 154:
```suggestion

      expect(code).toContain("import { Bar } from './somewhere'");
```

kaisermann (approved):

kaisermann:
@dummdidumm  LGTM, but I'm a bit out of context and busy with moving to another country, sorry for taking some days to check this! This is possible due to TS 4.5, right?

dummdidumm:
Correct. TS 4.5 introduced the new setting `preserveValueImports` along with DX to neatly separate types and values. You can read more on that here: https://devblogs.microsoft.com/typescript/announcing-typescript-4-5/#preserve-value-imports

package.json
```diff
@@ -87,7 +87,7 @@
     "sugarss": "^2.0.0",
     "svelte": "^3.42.0",
     "ts-jest": "^25.1.0",
-    "typescript": "^4.4.2"
+    "typescript": "^4.5.2"
```
kaisermann, line 90: Quick question: (when) have we removed the support for v3? 🤔

dummdidumm: never, these are the dev dependencies, not the peer dependencies.

PR #453, opened 2022-01-04
[sass] Tell, that we use legacy api

### LINKED ISSUE #452: [sass] preprocess mixes legacy and current api
**Describe the bug**
The [render and renderSync](https://sass-lang.com/documentation/js-api/modules#renderSync) methods have been marked as legacy. The svelte-preprocess expose [configuration options](https://github.com/sveltejs/svelte-preprocess/blob/f6ef334f45c1327e087d031bd6f3791970d7cd25/src/types/options.ts#L2) from new way.

From sass:
```ts
export function renderSync(options: LegacyOptions<'sync'>): LegacyResult;

export function render(
  options: LegacyOptions<'async'>,
  callback: (exception?: LegacyException, result?: LegacyResult) => void
): void;
```
new way:
```ts
export function compile(path: string, options?: Options<'sync'>): CompileResult;
```

bato3:
Partial-fix #452 

This patch only fixes the most pressing problem.

There is a question of how to deal with libraries that use legacy importers. 

### Tests

### DIFF
--- a/docs/preprocessing.md
+++ b/docs/preprocessing.md
@@ -354,7 +354,7 @@ The `scss/sass` preprocessor accepts the default sass options alongside two othe
 | `renderSync`     | `false`     | if `true`, use the sync render method which is faster for dart sass.                                                           |
 | `implementation` | `undefined` | pass the module to use to compile sass, if unspecified, `svelte-preprocess` will first look for `node-sass` and then for Sass. |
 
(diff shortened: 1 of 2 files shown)


dummdidumm:
The interface rename seems to have broken some tests

kaisermann:
This will be merged and used in #562. I actually tried to move to the new `compileString` API, but couldn't make it work in the testing environment unfortunately.

PR #473, opened 2022-02-01
fix: add # and numbers to ts store regex

### LINKED ISSUE #466: v4.9.0 onward invalid character error when using a dollar sign followed by a symbol (e.g. "$#")
**Describe the bug**
Writing a dollar sign '$' in code followed immediately by certain symbols (in my case I've confirmed it with '#' and '@') throws an invalid character error.

**Logs**
image

**To Reproduce**
Try setting a variable to a string the string "$#" and attempt to build the project or run dev and navigate to the respective page.
Here is a repo example: https://github.com/evan-sheehan/svelte-preprocess-bug
I used the "npm init svelte@next" command according to Sveltekit documentation and started a project using the demo template with Typescript by no Eslint and no Prettier.

**Expected behavior**
An error should not be thrown when svelte-preprocess parses '$#' in the context of a string or comment.

- Chrome Version 96.0.4664.45 (Official Build) (64-bit)
- Zorin 16 - Ubuntu 20.04. 3 LTS
- Svelte-preprocess anything 4.9.0 and after as far as I've tested.
- Using Sveltekit so Vite and Rollup are used internally.

andr3h3nriqu3s11:
Fixes #466

And also fixes another situation that led me to research this issue. Where something like this

```ts
let str = "$1$2";
```
or

```ts
let str = "$1a";
```
would result in a type error (TS1351) where it was assuming that `1$2` was a variable where it should not

### Tests

### DIFF
--- a/src/transformers/typescript.ts
+++ b/src/transformers/typescript.ts
@@ -153,7 +153,7 @@ function injectVarsToCode({
   // TS transformer (previous attemps have failed)
   const codestores = Array.from(
     contentForCodestores.match(
-      /\$[^\s();:,[\]{}.?!+\-=*/\\~|&%<>^`"'°§#1-9][^\s();:,[\]{}.?!+\-=*/\\~|&%<>^`"'°§#]*/g,
+      /\$[^\s();:,[\]{}.?!+\-=*/\\~|&%<>^`"'°§#0-9][^\s();:,[\]{}.?!+\-=*/\\~|&%<>^`"'°§#]*/g,
     ) || [],
     (name) => name.slice(1),
   ).filter((name) => !JAVASCRIPT_RESERVED_KEYWORD_SET.has(name));
--- a/src/transformers/typescript.ts
+++ b/src/transformers/typescript.ts
@@ -152,9 +152,7 @@ function injectVarsToCode({
   // TODO investigate if it's possible to achieve this with a
   // TS transformer (previous attemps have failed)
   const codestores = Array.from(
-    contentForCodestores.match(
-      /\$[^\s();:,[\]{}.?!+\-=*/\\~|&%<>^`"'°§#1-9]+/g,
-    ) || [],
+    contentForCodestores.match(/\$[a-zA-Z_][\w_$]*/g) || [],
     (name) => name.slice(1),
   ).filter((name) => !JAVASCRIPT_RESERVED_KEYWORD_SET.has(name));
 
--- a/src/transformers/typescript.ts
+++ b/src/transformers/typescript.ts
@@ -152,8 +152,9 @@ function injectVarsToCode({
   // TODO investigate if it's possible to achieve this with a
   // TS transformer (previous attemps have failed)
   const codestores = Array.from(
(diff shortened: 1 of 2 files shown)


andr3h3nriqu3s11:
I don't know if this is worth a new issue, but these errors also happen when the code is commented.

src/transformers/typescript.ts
```diff
@@ -152,9 +152,7 @@ function injectVarsToCode({
   // TODO investigate if it's possible to achieve this with a
   // TS transformer (previous attemps have failed)
   const codestores = Array.from(
-    contentForCodestores.match(
-      /\$[^\s();:,[\]{}.?!+\-=*/\\~|&%<>^`"'°§#1-9]+/g,
-    ) || [],
+    contentForCodestores.match(/\$[a-zA-Z_][\w_$]*/g) || [],
```
andr3h3nriqu3s11, line 155: This just makes more sense...
Matching for correct variables names instead of matching for things that are not variables

dummdidumm: Problem is that something like `$helloΩ` is a valid variable name but not matched by this regex.

src/transformers/typescript.ts
```diff
@@ -152,7 +152,9 @@ function injectVarsToCode({
   // TODO investigate if it's possible to achieve this with a
   // TS transformer (previous attemps have failed)
   const codestores = Array.from(
-    contentForCodestores.match(/\$[a-zA-Z_][\w_$]*/g) || [],
+    contentForCodestores.match(
+      /\$[^\s();:,[\]{}.?!+\-=*/\\~|&%<>^`"'°§#1-9][^\s();:,[\]{}.?!+\-=*/\\~|&%<>^`"'°§#]*/g,
```
andr3h3nriqu3s11, line 156: You need the groups because while the variable can not start with a number, it can have a number after it

src/transformers/typescript.ts
```diff
@@ -152,8 +152,9 @@ function injectVarsToCode({
   // TODO investigate if it's possible to achieve this with a
   // TS transformer (previous attemps have failed)
   const codestores = Array.from(
-    contentForCodestores.match(/\$[^\s();:,[\]{}.?!+\-=*/\\~|&%<>^`"'°§]+/g) ||
-      [],
+    contentForCodestores.match(
+      /\$[^\s();:,[\]{}.?!+\-=*/\\~|&%<>^`"'°§#1-9][^\s();:,[\]{}.?!+\-=*/\\~|&%<>^`"'°§#]*/g,
```
dummdidumm, line 156: Did you mean `0-9` instead of `1-9`?

```suggestion
      /\$[^\s();:,[\]{}.?!+\-=*/\\~|&%<>^`"'°§#0-9][^\s();:,[\]{}.?!+\-=*/\\~|&%<>^`"'°§#]*/g,
```

andr3h3nriqu3s11: I did

dummdidumm:
Thank you!

PR #487, opened 2022-02-23
fix: only try to load external files with relative paths

kaisermann:
Closes https://github.com/sveltejs/kit/issues/4062

The `isValidLocalPath` was checking the path validity the other way around (not remote). We should instead check if it's a valid local path (starts with `.` and doesn't contain interpolation).

### DIFF
--- a/src/modules/utils.ts
+++ b/src/modules/utils.ts
@@ -60,15 +60,8 @@ export async function hasDepInstalled(dep: string) {
   return (depCheckCache[dep] = result);
 }
 
-const REMOTE_SRC_PATTERN = /^(https?:)?\/\//;
-
 export function isValidLocalPath(path: string) {
-  return (
-    path.match(REMOTE_SRC_PATTERN) == null &&
-    // only literal strings allowed
-    !path.startsWith('{') &&
-    !path.endsWith('}')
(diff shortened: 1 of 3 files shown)


dummdidumm (approved):
Makes sense to me. I thought about whether or not this can be seen as a breaking change, but the existing behavior was unexpected, too, and the resolved file path in the `/app.js` case likely was wrong anyway, so I concluded that this is just a fix 👍

src/modules/utils.ts
```diff
@@ -60,15 +60,8 @@ export async function hasDepInstalled(dep: string) {
...
-const REMOTE_SRC_PATTERN = /^(https?:)?\/\//;
-
 export function isValidLocalPath(path: string) {
-  return (
-    path.match(REMOTE_SRC_PATTERN) == null &&
-    // only literal strings allowed
-    !path.startsWith('{') &&
-    !path.endsWith('}')
-  );
+  return path.startsWith('.') && !path.startsWith('{') && !path.endsWith('}');
```
Conduitry, line 64: Checking that it doesn't start with `{` is redundant now, as is checking that it doesn't end in `}` (probably).

PR #492, opened 2022-03-25
update package.json

gcxfd:
update package.json where I used 
I runned test , all ok


dummdidumm:
What was the motivation for doing this?
Also: we can't bump the peer dependencies, this is a breaking change strictly speaking. Therefore closing this - thank you.

PR #493, opened 2022-03-25
Coffee hack for label

gcxfd:

### Tests


gcxfd:
it work , i release a preview in @rmw/svelte-preprocess
```
yarn add  -D @rmw/svelte-preprocess @rmw/coffee-label-patch
```
but test can't pass , says can't found module , but I don't know how fix , typescript is to hard to use

mrkishi:
It's awesome that you figured out how to get that syntax to work!

Unfortunately `svelte-preprocess` can't use a patched Coffeescript compiler as that could break otherwise legitimate users' code.

I'm sure people looking to use reactive statements would use your patch, though.

gcxfd:
image
I modified the code so that the :label syntax is enabled only when the label:true argument is passed. So we don't break compile.
```
    svelte(
      preprocess: [
        sveltePreprocess(
          coffeescript: {
            label:true
            sourceMap: true
          }
          stylus: true
          pug: true
        )
      ]
    )
```

kaisermann:
Thanks for the contribution and the idea! However, as @mrkishi said, 

> Unfortunately svelte-preprocess can't use a patched Coffeescript compiler, at least not by default, as that would be quite surprising for users and would require on-going maintenance to make sure nothing breaks as the compiler changes.

Issue #495, opened 2022-04-01
Imports in less are added as dependencies with relative paths but webpack requires absolute paths


lundal:
**Describe the bug**
Imports in less are added as dependencies with relative paths ("node_modules/package/file.less"), but webpack requires absolute paths ("/home/user/project/node_modules/package/file.less").

**Logs**
```
WARNING in ./app/App.svelte
Invalid dependencies have been reported by plugins or loaders for this module. All reported dependencies need to be absolute paths.
Invalid dependencies may lead to broken watching and caching.
As best effort we try to convert all invalid values to absolute paths and converting globs into context dependencies, but this is deprecated behavior.
Loaders: Pass absolute paths to this.addDependency (existing files), this.addMissingDependency (not existing files), and this.addContextDependency (directories).
Plugins: Pass absolute paths to fileDependencies (existing files), missingDependencies (not existing files), and contextDependencies (directories).
Globs: They are not supported. Pass absolute path to the directory as context dependencies.
The following invalid values have been reported:
 * "node_modules/package/file.less"
 @ ./app/App.svelte
 @ ./app/app.js
```

**To Reproduce**
Use webpack with svelte-loader and svelte-preprocess:
```js
{
  loader: "svelte-loader",
  options: {
    preprocess: sveltePreprocess(),
  },
},
```
Use an import within less in a svelte file:
```svelte
<style lang="less">
  @import "package/file.less"
  ...
</style>
```
During build, webpack will complain about invalid dependencies.

Example that reproduces the error: https://github.com/lundal/svelte-less-example (`$ yarn install && yarn build`)

**Expected behavior**
No complaints from webpack

PR #496, opened 2022-04-01
fix: always use absolute paths for less imports

lundal:
Solves #495

Looks like the issue only affects imports from node_modules. Can't make a test without adding a dependency to the project.

### DIFF
--- a/src/transformers/less.ts
+++ b/src/transformers/less.ts
@@ -1,3 +1,5 @@
+import { isAbsolute, join } from 'path';
+
 import less from 'less';
 
 import { getIncludePaths } from '../modules/utils';
@@ -20,10 +22,14 @@ const transformer: Transformer<Options.Less> = async ({
     ...options,
   });
 
+  const dependencies = imports.map((path: string) =>
+    isAbsolute(path) ? path : join(process.cwd(), path),
+  );
+
   return {
     code: css,
     map,
-    dependencies: imports,
+    dependencies,
   };
 };
 


src/transformers/less.ts
```diff
@@ -20,10 +22,14 @@ const transformer: Transformer<Options.Less> = async ({
     ...options,
   });
 
+  const dependencies = imports.map((path: string) =>
+    isAbsolute(path) ? path : join(process.cwd(), path),
```
kaisermann, line 26: Are the paths relative to the cwd? Something like `node_modules/...`?

lundal: As described in #495 you currently a warning like this:
```
WARNING in ./app/App.svelte
Invalid dependencies have been reported by plugins or loaders for this module. All reported dependencies need to be absolute paths.
Invalid dependencies may lead to broken watching and caching.
As best effort we try to convert all invalid values to absolute paths and converting globs into context dependencies, but this is deprecated behavior.
Loaders: Pass absolute paths to this.addDependency (existing files), this.addMissingDependency (not existing files), and this.addContextDependency (directories).
Plugins: Pass absolute paths to fileDependencies (existing files), missingDependencies (not existing files), and contextDependencies (directories).
Globs: They are not supported. Pass absolute path to the directory as context dependencies.
The following invalid values have been reported:
 * "node_modules/package/file.less"
 @ ./app/App.svelte
 @ ./app/app.js
```
As you can see, the path is `node_modules/package/file.less` but should be `/home/user/project/node_modules/package/file.less`.

I based the fix on this code in `src/transformers/scss.ts`:
```ts
  const absoluteEntryPath = isAbsolute(result.stats.entry)
    ? result.stats.entry
    : join(process.cwd(), result.stats.entry);
```
Works for https://github.com/lundal/svelte-less-example atleast.

kaisermann (approved):

lundal:
Any idea when this will make it into a release, @kaisermann ?

kaisermann:
@lundal Released in `v4.10.5` 🎉

lundal:
Thank you for following up on this so quickly :zap:

PR #500, opened 2022-04-08
fix: paths in postcss sourcemap sources array (Sourcemap ... points to missing source files)

bfanger:
When using Vite's  new [devSourcemap](https://vitejs.dev/config/#css-devsourcemap) you'll be getting warnings like:

```
Sourcemap for ".../src/routes/__layout.svelte" points to missing source files
```

Example SvelteKit project: https://github.com/bfanger/svelte-project-template 

```js
// svelte.config.js
/** @type {import('@sveltejs/kit').Config} */
export default {
  preprocess: preprocess({ sourceMap: true }),
  kit: {
    vite: { css: { devSourcemap: true } },
  },
};
```
When no `to:` is passed to `postcss()`, PostCSS will use NodeJS current working directory as fallback, which generates  invalid paths. 

This PR adds the `to` parameter, which allows PostCSS to generate the correct relative paths into the sourcemap sources array.

### DIFF
--- a/src/transformers/globalStyle.ts
+++ b/src/transformers/globalStyle.ts
@@ -75,6 +75,7 @@ const transformer: Transformer<Options.GlobalStyle> = async ({
 
   const { css, map: newMap } = await postcss(plugins).process(content, {
     from: filename,
+    to: filename,
     map: options?.sourceMap ? { prev: map } : false,
   });
 
--- a/src/transformers/postcss.ts
+++ b/src/transformers/postcss.ts
@@ -15,6 +15,7 @@ async function process({
 }) {
   const { css, map, messages } = await postcss(plugins).process(content, {
     from: filename,
+    to: filename,
     map: { prev: sourceMap, inline: false },
     parser,
     syntax,


dominikg (approved):

floehopper:
I've started seeing these warnings having just upgraded from v1.0.0-next.298 -> next.310, but I don't have `devSourcemap` enabled. Will this also be fixed by these changes?

bfanger:
@floehopper Dig into your local node_modules/svelte-preprocess
Edit both node_modules/svelte-preprocess/dist/transformers/**globalStyle.js** & **postcss.js**

and replace:
```
from: filename,
```
with
```
from: filename, 
to: filename,
```
And see if that fixes it, if not then there is something else also generating invalid paths

floehopper:
@bfanger Yes, that fixes them - thanks! 👍

Gin-Quin:
I'm having this issue for Stylus files. I fixed it by applying `path.resolve` to all sourcemap sources in `transformers/stylus.ts`. I created a PR for that.

PR #517, opened 2022-05-11
chore: switch to pnpm

mcmxcdev:
## Description
I made sure to run `pnpm import` to avoid updating subdependencies with a fresh lock file install.

I had to bump the minimum node v14 requirement and install missing `@types/babel__core` so that it works with pnpm 7.
I updated some GitHub Actions to v2, but they should be within node v14 requirement still.

I didn't touch the example apps, should we replace them with pnpm as well? I tried to test run the install, but ran into issues with non-existing python2 on my system (Probably due to deprecated node-sass).

`svelte-rollup` has both npm and yarn lock file (which is definitely wrong) and `sapper-rollup` has npm lock file currently.

## Changes
- remove yarn-specific `.gitignore` entries
- update `volta` node version to meet requirements from pnpm 7
- update npm scripts to run with pnpm
- add `@types/babel__core` type definitions
- update pull request template
- update CI GitHub Action to use pnpm

### Tests


antony:
I'm in favour of this. I vote to leave the example as npm for now, since npm is more common than pnpm.

antony (approved):

PR #518, opened 2022-05-15
fix: add mixin const (#467)

fedorovvvv:
Problem with const mixin in pug
https://github.com/sveltejs/svelte-preprocess/discussions/467


kaisermann (approved):
Nice!

PR #520, opened 2022-05-25
feat: support postcss-load-config 4

### LINKED ISSUE #519: Support postcss-load-config 4

bluwy:
Fix #519

Checking the code changes in postcss-load-config, this should be enough to suppress peer dep issues.

### Tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -118,7 +118,7 @@
     "coffeescript": "^2.5.1",
     "less": "^3.11.3 || ^4.0.0",
     "postcss": "^7 || ^8",
-    "postcss-load-config": "^2.1.0 || ^3.0.0",
+    "postcss-load-config": "^2.1.0 || ^3.0.0 || ^4.0.0",
     "pug": "^3.0.0",
     "sass": "^1.26.8",
     "stylus": "^0.55.0",


package.json
```diff
@@ -118,7 +118,7 @@
     "coffeescript": "^2.5.1",
     "less": "^3.11.3 || ^4.0.0",
     "postcss": "^7 || ^8",
-    "postcss-load-config": "^2.1.0 || ^3.0.0",
+    "postcss-load-config": "^2.1.0 || ^3.0.0 || ^4.0.0",
```
baseplate-admin, line 121: Hi there,

I am a bit confused about this change. 

In my opinion the change should be like this.
```
"postcss-load-config": "^2.1.0 
```

Because in current state we are matching everything `v3.0.0` and `v4.0.0` which then can again break in version 5.0.0.

So in my opinion matching everything above `v2.1.0` is the best approach :)

Thanks for reading :D

bluwy: I guess you mean `>=2.1.0` according to semver, as `^2.1.0` only matches anything version 2 (version 3 is out of scope). I think either way is fine and it's up to the maintainers preference, as there are pros/cons for both approach

baseplate-admin: > I guess you mean >=2.1.0 according to semver, as ^2.1.0 only matches anything version 2 (version 3 is out of scope). I think either way is fine and it's up to the maintainers preference, as there are pros/cons for both approach

Forgive my mistake and that is exactly what i mean :D

Conduitry: Yeah, this feels like a trade-off to me, and I don't have a strong opinion either way. If you only mention version 2, 3, and 4, then you need to manually add 5 once that's released and you make sure that its breaking changes don't affect you. If you use `>=2.1.0`, you're asserting that this will work regardless of any breaking changes `postcss-load-config` might make in the future, which might be fine for a while, until one day it bites some user in potentially a really confusing way.

baseplate-admin: @Conduitry Thanks for giving your opinion on this matter :)

Have a good day :)

el3um4s (approved):

dummdidumm (approved):

PR #527, opened 2022-06-13
Fix link to preprocess API docs

JoelHobson:
I think the URL for the preprocess API docs changed at some point. Clicking it takes me to the top of the docs page. I've replaced it with what I believe to be the correct URL.

Apologies for not creating an issue for this, it didn't strike me as providing value for such a small change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@
 
 ## What is it?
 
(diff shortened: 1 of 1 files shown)


bluwy (approved):

dummdidumm (approved):
Thanks! Opening an issue for this first would indeed be overkill, all good 👍

PR #531, opened 2022-06-16
Added example for alias importer

grbd:
Hi,
I've added an example of using the importer callback within svelte-preprocess.
I was inspired to write it based on the example here - https://github.com/sveltejs/svelte-preprocess/issues/97

This solved a problem for me when trying to get vite alias's working inside the svele:head section of a component
It took me a while to figure out so I thought it might be worth documenting.

### DIFF
--- a/docs/alias-importer.md
+++ b/docs/alias-importer.md
@@ -1,0 +1,180 @@
+# Alias Importer Example
+
+One of the features of the vite build sytem is the aliasing of import paths.
+For example, you can provide a shortcut to the path of a css file in an import statement instead of the full path.
+
+For example:
+```html
+<script>
+  import "@styles/example1.css";
+</script>
+
+<style lang="scss">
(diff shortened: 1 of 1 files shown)


dummdidumm:
Wow, that's a very thorough and good explanation of what to do - thank you! Some of these things feel a little unrelated to `svelte-preprocess` (mentioning the Vite options), but I also don't know whereelse the best place is to put this. Ultimately this feels like something that belongs into a section of guidelines on the SvelteKit site, which doesn't exist yet, so this may be the next best place in the meantime.

bluwy:
I also think it's better if this is published as a blog. It should be more easily indexed by search engines too. It's very specific to a usecase which I'm not sure fit in the general documentation.

grbd:
Ok I'll post this as part of a blog entry once I get my blog up and running

Issue #504, opened 2022-04-14
Version 5 (major release) planning


dummdidumm:
This issue is intended to collect breaking changes we want to do for the next major version of `svelte-preprocess`. What currently comes to my mind:

- remove support for the default language (deprecated already)
- remove support for `lang="typescript"` and `type="text/..."` (deprecated already AFAIK)
- remove support for Node 10 (EOL) and 12 (EOL end of April)
- remove peer dependencies and instead tell in the docs what versions we tested it with. Npm 7 automatically installs peer dependencies, which means useless downloads for almost everyone. UPDATE: There's https://docs.npmjs.com/cli/v8/configuring-npm/package-json#peerdependenciesmeta so we can keep peer deps. We can do this now already.
- Only support the new SASS API (right now we only support the legacy API). This means bumping the peer dep.

mcmxcdev:
You might want to add another bullet point to consider switching from yarn v1, which has frozen development for a while, to e.g. pnpm (which would be in line with the package manager choice of other sveltejs repos)

kaisermann:
> This issue is intended to collect breaking changes we want to do for the next major version of svelte-preprocess. What currently comes to my mind:

I was hoping v5 would be released at a similar time as the RFCed new preprocess API, but it seems that's still far from happening. Removing deprecations and support for EOL node versions is a good start! We don't need to make it a very big breaking change 🤷

> You might want to add another bullet point to consider switching from yarn v1, which has frozen development for a while, to e.g. pnpm (which would be in line with the package manager choice of other sveltejs repos). 

I have no qualms with switching to another package manager, even more, if it's to the same one used by other Svelte projects.

frangio:
Peer dependencies can be marked optional as shown here: https://docs.npmjs.com/cli/v8/configuring-npm/package-json#peerdependenciesmeta

I believe this turns off the automatic installs.

PR #542, opened 2022-07-21
docs(updated link to compiler options for scss & sass)

aarondiel:
hint for where to find additional sass / scss compiler options.

this was very confusing for me, since the docs suggest that you can pass **importers** to the sass preprocessor and gave a different type definition from what svelte-preprocess expects.

taking the following code snippet from the [current sass docs](https://sass-lang.com/documentation/js-api/interfaces/FileImporter) does not work:
```javascript
preprocess({
  scss: {
    importers: [{
      // An importer that redirects relative URLs starting with "~" to
      // `node_modules`.
      findFileUrl(url) {
        if (!url.startsWith('~')) return null;
        return new URL(url.substring(1), pathToFileURL('node_modules'));
      }
    }]
  }
})
```

while something like this would:
```javascript
preprocess({
  scss: {
    importer: (url) => {
      console.log(url)

      return {
        file: url
      }
    }]
  }
})
```

i hope this is only a temporary fix for when svelte-preprocess moves to a newer sass version


dummdidumm:
Thank you! I added "move to new SASS API" to the list of the next major release planning (#504)

PR #557, opened 2022-11-29
add tailwind-as-html preprocessor

Temperantia:
Enables using tailwind classes as html attributes

image


Conduitry:
Are there any other Tailwind integrations out there that support or encourage this?

I have my doubts about this feature, but if we were to add it, it would need to be in a way that doesn't add `cheerio` as a prod dependency for every user. It would also need to be documented, and preferably would also be opt-in.

Temperantia:
No, it's only an idea I had a few days ago, I started using it for my own code, it feels a lot more convenient:
- "class" is no longer needed
- Elements become vertical thanks to the formatter instead of a tedious long line
- No need for class name builders
- Conditional classes can be done in Svelte

I can work on removing cheerio and document a few lines.
I'm not sure what you mean by otp-in, it relies on using `<template lang="tailwind">`

Let me know if that makes sense

kaisermann:
Hey, @Temperantia 👋  thanks for your idea! We already have some ways of using `tailwind` that are a bit more conventional (like [this one](https://tailwindcss.com/docs/guides/sveltekit) or even [this other one](https://windicss.org/integrations/svelte.html)). `svelte-preprocess` tries to stay close to the native transformers, so your suggestion feels like a great case for a custom preprocessor.

PR #560, opened 2022-12-10
Upgrade a couple dependencies

benmccann:
`magic-string` was causing a deprecation warning
`node-sass` wouldn't install on my machine because it was too old
I was getting some kind of error from husky. I'm not sure if upgrading it helped. I had to delete the precommit hooks to send this


Conduitry:
For future reference, you can commit with the `-n` (`--no-verify`) flag to skip precommit hooks.

benmccann:
this repo has really bit rotted :stuck_out_tongue_closed_eyes: closing in favor of https://github.com/sveltejs/svelte-preprocess/pull/561

PR #561, opened 2022-12-10
Drop Node 10 and 12. Add 16 and 18

benmccann:
Might require a major version bump? Is the changelog just created manually?

The CI is failing on it due to outdated Node versions. This should fix it
`node-sass` wouldn't install because it was too old
`magic-string` was causing a deprecation warning
I was getting an error from husky. I'm not sure if upgrading it helped, but hopefully it doesn't hurt


kaisermann:
Hey, @benmccann 👋 Thanks! I will take care of this! The changelog is written automatically when publishing via the `conventional-changelog` CLI. 

> Might require a major version bump?

Yeah, it does. Better to be on the safe side here.

kaisermann:
I'm getting some weird, possibly `node-gyp` related, when trying to install the project's deps with `node-sass@8`. Given that changing the node requirement already implies in a breaking change, we could look into just supporting `sass` and removing support for `node-sass` and some point.

(deprecation notice for `node-sass`: https://github.com/sass/node-sass/issues/2952)

Issue #508, opened 2022-04-28
Conflicting peer dependency with postcss-mixins (sugarss) 


Azarattum:
**Describe the bug**
Unable to add `postcss-mixins` to a svelte project.

**Logs**
```
npm ERR! code ERESOLVE
npm ERR! ERESOLVE could not resolve
npm ERR!
npm ERR! While resolving: svelte-preprocess@4.10.6
npm ERR! Found: sugarss@4.0.1
npm ERR! node_modules/sugarss
npm ERR!   sugarss@"^4.0.1" from postcss-mixins@9.0.2
npm ERR!   node_modules/postcss-mixins
npm ERR!     dev postcss-mixins@"^9.0.2" from the root project
npm ERR!
npm ERR! Could not resolve dependency:
npm ERR! peerOptional sugarss@"^2.0.0" from svelte-preprocess@4.10.6
npm ERR! node_modules/svelte-preprocess
npm ERR!   dev svelte-preprocess@"^4.10.1" from the root project
npm ERR!   svelte-preprocess@"^4.0.0" from svelte-check@2.7.0
npm ERR!   node_modules/svelte-check
npm ERR!     dev svelte-check@"^2.2.6" from the root project
npm ERR!
npm ERR! Conflicting peer dependency: sugarss@2.0.0
npm ERR! node_modules/sugarss
npm ERR!   peerOptional sugarss@"^2.0.0" from svelte-preprocess@4.10.6
npm ERR!   node_modules/svelte-preprocess
npm ERR!     dev svelte-preprocess@"^4.10.1" from the root project
npm ERR!     svelte-preprocess@"^4.0.0" from svelte-check@2.7.0
npm ERR!     node_modules/svelte-check
npm ERR!       dev svelte-check@"^2.2.6" from the root project
npm ERR!
npm ERR! Fix the upstream dependency conflict, or retry
npm ERR! this command with --force, or --legacy-peer-deps
npm ERR! to accept an incorrect (and potentially broken) dependency resolution.
```

**To Reproduce**
- `npm i postcss-mixins -D` in a project with `svelte-preprocess`

**Expected behavior**
Install without a problem.

**Information about your project**

- Your operating system: Windows 10
- `svelte-preprocess` version: 4.10.6
- Project uses SvelteKit

**Possible solution**
Update `sugarss` to `^4.0.1`.

dummdidumm:
[...]

Issue #485, opened 2022-02-19
[idea] Drop node-sass support


bfanger:
**Is your feature request related to a problem? Please describe.**
`yarn install` fails on modern node versions.

https://sass-lang.com/blog/libsass-is-deprecated

**Describe the solution you'd like**
'sass' (dart-sass) is the recommended successor.
Which (aside from additional slash-div warnings) has been a nice experience
(didn't break when upgrading nodejs and is faster than ruby sass)

**Describe alternatives you've considered**
`nvm` & `docker run node:12`

**How important is this feature to you?**
Not important, I'm not a svelte-preprocess maintainer 😅

**Additional context**
I want to fix a bug in svelte-process, but had to jump through hoops to get `yarn dev` working.
Other potential contributors might have given up more easily.

kaisermann:
Hey @bfanger 👋  sorry for the delay! This will be handled by #562

PR #562, opened 2022-12-10
v5

### LINKED ISSUE #504: Version 5 (major release) planning

kaisermann:
Closes #504, #485, #452, #508. This PR basically does what is described in #504:

- Bumps minimum node version from 10 to 14
- Remove `node-sass` support. `libSass` (aka `node-sass`) was [deprecated a while ago.](https://sass-lang.com/blog/libsass-is-deprecated). Use [`sass`](https://www.npmjs.com/package/sass) instead.
- Remove support for the `type="text/..."` attribute (was deprecated)
- Remove support for custom default languages `.defaults`  (was deprecated)

From #504 (cc @dummdidumm)
> Only support the new SASS API (right now we only support the legacy API). This means bumping the peer dep.

This will be the only thing not implemented right now as I'm not managing to make it work in the test environment. There's also a problem with sync/async importers which I'm not sure how to deal with just yet.


kaisermann:
Updating to `sass` only and replacing the `render` call with `compileString` is harder than I thought 😅

kaisermann:
@dummdidumm The result of the `typescript` transformer still returns a `diagnostics` list alongside the rest of the `Processed` result. Is this used somewhere within the language server or in any other tool? Otherwise, I think we could remove this

src/transformers/scss.ts
```diff
@@ -59,24 +38,17 @@ const transformer: Transformer<Options.Sass> = async ({
   filename,
   options = {},
 }) => {
-  let implementation = options?.implementation ?? sass;
```
kaisermann, line 62: Since `node-sass` is deprecated, I think we might as well remove support for it altogether.

CHANGELOG.md
```diff
@@ -1,3 +1,38 @@
...
+
+### Features
+
+* 🎸 bump minimum node version to 14 ([30a9287](https://github.com/sveltejs/svelte-preprocess/commit/30a9287c474241ef0bf03bade04404e120cb1ec7))
+* drop Node 10 and 12. Add 16 and 18 ([#561](https://github.com/sveltejs/svelte-preprocess/issues/561)) ([55c051d](https://github.com/sveltejs/svelte-preprocess/commit/55c051d6a63e5f7da430c736b2384995c0ec706b))
+
+
+### Performance Improvements
+
+* ⚡️ remove support for deprecated node-sass ([4ae1ff1](https://github.com/sveltejs/svelte-preprocess/commit/4ae1ff1db17ff98d97be1a0ec538ef40d8affdcb))
```
benmccann, line 22: Might not need this line since it's also listed under breaking changes

benmccann (approved):

kaisermann:
@benmccann Just released v5 🎉  those installation errors should stop now 🙏

dummdidumm:
> @dummdidumm The result of the `typescript` transformer still returns a `diagnostics` list alongside the rest of the `Processed` result. Is this used somewhere within the language server or in any other tool? Otherwise, I think we could remove this

Nope, not used, so can be removed.
Thank you for releasing the next major 🎉

benmccann:
thanks!!

PR #564, opened 2022-12-16
docs: Update documentation about Sass options

### LINKED ISSUE #563: Doc out of date: Sass renderSync option no longer does anything
**Describe the bug**

This bit of documentation is out of date:
https://github.com/sveltejs/svelte-preprocess/blob/4218419a145c29248d682cddc3eda557409ec38b/docs/preprocessing.md?plain=true#L356

Following https://github.com/sveltejs/svelte-preprocess/commit/06fd5b9af92c985cbf1e7a89e5d576102b66786b it seems like svelte-preprocess can *only* use Sass's `renderSync` API, so the option no longer has an effect.

This part of the doc might cause users to think (incorrectly) that they are using Sass's `render` API unless they specify the option.

**Logs**
N/A

**To Reproduce**
Read https://github.com/sveltejs/svelte-preprocess/blob/4218419a145c29248d682cddc3eda557409ec38b/docs/preprocessing.md#scss-sass

**Expected behavior**
Behavior described by doc matches behavior of code.

**Stacktraces**
N/A

N/A

**Additional context**
N/A

alecglassford:
Fixes #563: The doc had incorrect/outdated information about the `renderSync` option). The update notes that `renderSync` is always used as of svelte-preprocess 5.0.0.

While addressing this, I also added a more specific cross reference link for users to see which options can be passed: https://sass-lang.com/documentation/js-api/interfaces/LegacyStringOptions

I also removed the table, since that formatting now longer seemed useful, and made other minor wording adjustments to this section.

### Tests

### DIFF
--- a/docs/preprocessing.md
+++ b/docs/preprocessing.md
@@ -349,12 +349,9 @@ input(on:input="{bar}")
 
 ### scss, sass
 
-The `scss/sass` preprocessor accepts the default sass options alongside two other props:
+The `scss/sass` preprocessor accepts the options listed in the [Sass `LegacyStringOptions` API reference](https://sass-lang.com/documentation/js-api/interfaces/LegacyStringOptions) with some exceptions: the `file` and `data` properties are not supported. Instead, use the `prependData` property if you want to prepend some content to your `scss` content.
 
-| Option           | Default     | Description                                                                                                                    |
-| ---------------- | ----------- | ------------------------------------------------------------------------------------------------------------------------------ |
-| `renderSync`     | `false`     | if `true`, use the sync render method which is faster for dart sass.                                                           |
-You can check the [Sass Legacy API reference](https://sass-lang.com/documentation/js-api#legacy-api) for specific Sass options. The `file` and `data` properties are not supported. Instead, use the `prependData` property if you want to prepend some content to your `scss` content.
+Note: `svelte-preprocess` (version 5.0.0 and later) always uses [Sass's legacy `renderSync` API](https://sass-lang.com/documentation/js-api/#legacy-api).
 
 Note: `svelte-preprocess` automatically configures inclusion paths for your root directory, `node_modules` and for the current file's directory.
 


docs/preprocessing.md
```diff
@@ -349,12 +349,9 @@ input(on:input="{bar}")
 
 ### scss, sass
 
-The `scss/sass` preprocessor accepts the default sass options alongside two other props:
+The `scss/sass` preprocessor accepts the options listed in the [Sass `LegacyStringOptions` API reference](https://sass-lang.com/documentation/js-api/interfaces/LegacyStringOptions) with some exceptions: the `file` and `data` properties are not supported. Instead, use the `prependData` property if you want to prepend some content to your `scss` content.
```
alecglassford, line 352: Wasn't sure if

> ... with some exceptions: the `file and `data` properties are not supported. Instead, use the `prependData` property if you want to prepend some content to your `scss` content.

should perhaps be changed to something more like

> 
> ... with some exceptions. The following options are not supported, as `svelte-preprocess` overrides them:
> 
> * `includePaths`
> * `sourceMap`
> * `sourceMapEmbed`
> * `omitSourceMapUrl`
> * `outFile`
> * `outputStyle`
> * `file`
> * `data`
> * `importer`
> 
> If you want to prepend some content to your `scss`, you can use `prependData` option that's specific to `svelte-preprocess`.

Just because it looks like all these options get clobbered: https://github.com/sveltejs/svelte-preprocess/blob/4218419a145c29248d682cddc3eda557409ec38b/src/transformers/scss.ts#L46-L62

I'm not really familiar with this codebase though, so perhaps I've misunderstood, so I'll leave it up to the maintainers if they want to add this level of additional detail.

kaisermann: I don't think omitting these is a problem. as most don't make sense in the context of a preprocessor for the style content.

Explaning each of them:
`includePaths` can be passed (as you can see from the `options.includePaths` on line 46. Most of the other props are not that relevant given the context sass is being used. We need to force source maps, so those should be set regardless. IIRC, the `outFile` must be set due to a limitation of how sass returns the imported dependencies. If there's no `outFile`, the dependency list comes empty. `outputStyle` will not minify the output, which I think should be expected in this case.

kaisermann (approved):

kaisermann:
Hey @alecglassford 👋  thanks for catching that! 🎉

PR #571, opened 2023-01-20
update sorcery

Rich-Harris:
see https://github.com/sveltejs/kit/issues/8611


dummdidumm:
seems like the import changed from default to named for some reason ..? Also, it seems that `sorcery` inlines the sourcemap-codec package while building, so that dependency should probably be a devDependency..?

PR #572, opened 2023-02-04
fix: allow pug template inheritance

### LINKED ISSUE #459: cannot use pug template inheritance
## bug

cannot use pug template inheritance

## test

sample repo: https://github.com/benblazak/_svelte_preprocess__pug_template_inheritance

```bash
git clone https://github.com/benblazak/_svelte_preprocess__pug_template_inheritance.git
cd _svelte_preprocess__pug_template_inheritance
npm install
npm run dev -- --open
```

- try the mixin link
  - this should work
- try the template link
  - this should fail with the error message
    > Declaration of template inheritance ("extends") should be the first thing in the file. There can only be one extends statement per file.

## ideas

i think the reason `extends ...` is not the first thing in the file is because the svelte mixins are prepended here: https://github.com/sveltejs/svelte-preprocess/blob/main/src/transformers/pug.ts#L68

benblazak:
Fixes #459.

Allows for pug template inheritance.

The issue was that `extends` statements must occur first in a pug file, but the svelte mixins were being inserted at the top of the file contents before being passed to pug.

I made changes so that if an `extends` statement occurs as the first line in a file, it is split off from the rest of the content and inserted before the svelte mixins.

- [x] test
- [x] lint

## notes

I wasn't able to research or test whether there are other things that should also be kept before svelte mixins.


src/transformers/pug.ts
```diff
@@ -67,8 +67,21 @@ const transformer: Transformer<Options.Pug> = async ({
     ...options,
   };
 
+  // separate content that needs to come before the svelte mixins
+  let preContent = '';
+
+  if (content.startsWith('extends')) {
```
kaisermann, line 73: Given the [example of inheritance from pug's documentation](https://pugjs.org/language/inheritance.html), what do you think of also adding support for files that start with a comment like:

```pug
//- page-a.pug
extends layout.pug

block scripts
  script(src='/jquery.js')
  script(src='/pets.js')
```

kaisermann:
Hey @benblazak, thanks for this! Just added a small suggestion 🙏

benblazak:
Ahh good catch @kaisermann. To properly allow comments though, we'd probably need to parse them (there can be more than one before `extends` right? and multi-line comments are allowed in that location too?), which is more than I'm up for right now (in the middle of a move -> new job).

Thinking about it, it might make more sense to stop including mixins by default at all, and put them in a separate file for people to include if they want. But that would be a breaking change.

simiraaaa:
@kaisermann @benblazak 
https://github.com/sveltejs/svelte-preprocess/pull/599

Fixed so that extends works with comments and `prependData`.
Please refer to it.

(Machine translation from Japanese to English at DeepL)

PR #583, opened 2023-03-05
fix: remove deprecated package @types/sass

### LINKED ISSUE #582: @types/sass deprecated
**Describe the bug**
Get a deprecated warning when creating a new project with Sveltekit

**Logs**
` WARN  deprecated @types/sass@1.45.0: This is a stub types definition. sass provides its own type definitions, so you do not need this installed. `

```bash
$> pnpm why @types/sass
Legend: production dependency, optional only, dev only

my-project@0.0.1 /path/to/my-project

devDependencies:
svelte-check 3.0.4
└─┬ svelte-preprocess 5.0.1
  └── @types/sass 1.45.0
```

tlaundal:
Fixes #582.

My best guess is that `@types/sass` was included in `dependencies` (and not `devDependencies`) as a convenience for users. With types now being included in the main `sass` package, it should no longer be an issue that people might have `sass` without the types, so we can safely remove the types here without promoting `sass` from `devDependencies` to `dependencies`. 

This removes the warning described in #582.

### Tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -89,7 +89,6 @@
   },
   "dependencies": {
     "@types/pug": "^2.0.6",
-    "@types/sass": "^1.43.1",
     "detect-indent": "^6.1.0",
     "magic-string": "^0.27.0",
     "sorcery": "^0.11.0",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -11,7 +11,6 @@ specifiers:
   '@types/jest': ^27.5.2
   '@types/node': ^14.18.34
   '@types/pug': ^2.0.6
-  '@types/sass': ^1.43.1
   '@types/stylus': ^0.48.38
   autoprefixer: ^9.8.8
   babel-minify: ^0.5.2
@@ -40,7 +39,6 @@ specifiers:
 
 dependencies:
   '@types/pug': 2.0.6
-  '@types/sass': 1.43.1
   detect-indent: 6.1.0
   magic-string: 0.27.0
   sorcery: 0.11.0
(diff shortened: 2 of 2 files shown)


Conduitry:
This is probably a good change to make, but it should probably also accompany a bump of the peerdep on Sass (to when the types are included in the package), which would necessitate a major bump - so we should try to see what other breaking changes we should batch that with.

tlaundal:
Hmm, fair point. We require version `^1.26.8` as peer, while the types were added  in [version `1.45.0`](https://github.com/sass/dart-sass/blob/main/CHANGELOG.md#1450).

I'm not quite sure if type declarations are resolved in the same way as normal modules, but if they are then including `@types/sass` as a dependency of this package is not enough to ensure it is available for the users of the package. This is likely a moot point since it probably works in 90% of cases, though.

Would you like me to add the bump for the peer dep in this PR?

dummdidumm:
I don't see any mention of `@types/sass` anywhere in the code base, so I'm wondering if this is just a leftover. @kaisermann do you have more info on why this is added as a dependency?

tlaundal:
> I don't see any mention of `@types/sass` anywhere in the code base, so I'm wondering if this is just a leftover and we can just remove it. @kaisermann do you have more info on why this is added as a dependency?

Any import of sass would use these types, at least before sass included its own types. At least some of them show up with a [search](
https://github.com/search?q=repo%3Asveltejs%2Fsvelte-preprocess+%22from+%27sass%27%22&type=code).

I can't speak for the original intention, but the dependency was introduced [here](https://github.com/sveltejs/svelte-preprocess/pull/80/commits/bf685f5710b28c45c5e6cfdae732d47ab71aba32#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R82), and it seems very much deliberate that only a few of the new types were included as dependencies, with the rest being only dev dependencies. My best guess is it was to ensure the types exported by this package would be valid in projects that import it.

This is also the breakage Conduitry is referencing, that the type check for a project would break because they are on a version of sass that doesn't export the types, and they are no longer included as a dependency by svelte-preprocess.

kaisermann:
>  @kaisermann do you have more info on why this is added as a dependency?

As @tlaundal rightly said, it was for the consumer's convenience as the options object would be typed, etc. 

Let me investigate what could happen if we just removed it.

kaisermann (approved):

kaisermann:
Released in `v5.0.2`

PR #622, opened 2023-12-17
breaking: remove preserve option

benmccann:
closes https://github.com/sveltejs/svelte-preprocess/issues/305
closes https://github.com/sveltejs/svelte-preprocess/issues/312

Turns out the `preserve` option doesn't actually do anything useful anymore because it's now looking at `lang` and not `type` as it was originally. The original issues are already fixed without specifying `preserve` at all.


test/autoProcess/autoProcess.test.ts
```diff
@@ -106,20 +106,14 @@ describe('options', () => {
...
-  it('should NOT preprocess preserved languages', async () => {
-    const input = `<div></div><script lang="ld+json">{"json":true}</script>`;
-    const opts = sveltePreprocess({
-      preserve: ['ld+json'],
-      aliases: [['ld+json', 'structuredData']],
-      structuredData() {
-        return { code: '', map: '' };
-      },
-    });
+  it('should NOT preprocess unrecognized languages', async () => {
```
kaisermann, line 109: It seems that I forgot how the auto preprocessor is working 😅  I didn't remember it's skipping unknown languages 🤔

benmccann: At some point in the past it wasn't, but it's nice that it is now. Saved me from having to implement it!

I think it's not really that it's skipping unknown languages but that it's looking at the `lang` field, so you can specify whatever you want in `type` since it's not using that field. Then since this doesn't specify a `lang` it gets skipped.

dummdidumm (approved):

PR #632, opened 2024-05-09
fix: Adjust globalifySelector to not split selectors with parentheses.

### LINKED ISSUE #501: Global styles not supporting composed selectors like :is()
**Problem**

Using the `global` attribute on a style block will prevent any use of composed CSS selectors inside it. Markup as simple as the following will trigger the error.
```svelte
<style global>
    article :is(h1, h2) {
        font-family: "EB Garamond";
    }
</style>
```

**Output**
```
<redacted>/src/layouts/article.svelte:47:21 :global(...) must contain a single selector
 45 |      }
 46 |  
 47 |      :global(article) :global(:is(h1), :global(h2)) {
                            ^
 48 |          font-family: "EB Garamond";
 49 |      }
```

I would expect `:global(article h1) :global(article h2)` to be a more correct preprocessed output.

gregschmit:
Fixes #501

I ran into an issue where I could not use simple tailwind apply directives for dark mode and also couldn't import Daisy UI. I tracked it down to a problem with `svelte-preprocess` creating `:global` rules that svelte is not happy with. User https://github.com/MagDevX suggested an adjustment to `globalifySelector.js` `combinatorPattern` to prevent selectors with parenthesis from being split.

### DIFF
--- a/src/modules/globalifySelector.ts
+++ b/src/modules/globalifySelector.ts
@@ -6,7 +6,7 @@
  * escaped combinators like `\~`.
  */
 // TODO: maybe replace this ugly pattern with an actual selector parser? (https://github.com/leaverou/parsel, 2kb)
-const combinatorPattern = /(?<!\\)(?:\\\\)*([ >+~,]\s*)(?![^[]+\]|\d)/g;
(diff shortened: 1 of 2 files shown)


gregschmit:
Update: I added a test that shows the problem and it fails in `main` but passes on this branch. I ran tests and everything passed.

dummdidumm (approved):
Thank you!

PR #638, opened 2024-06-11
chore: upgrade to pnpm 9 and drop Node 16 support

benmccann:
To match `@sveltejs/svelte` and `@sveltejs/kit`


dummdidumm:
Since it's breaking we can only do it in a major

kaisermann (approved):

PR #640, opened 2024-06-12
breaking: svelte-preprocess 6

### LINKED ISSUE #488: TypeScript not being processed when no filename is given
**Describe the bug**
`if (filename == null) return { code: content };`
in [svelte-preprocess/src/transformers/typescript.ts](https://github.com/sveltejs/svelte-preprocess/blob/main/src/transformers/typescript.ts) line 497 is the issue.

**To Reproduce**
Call `svelte.preprocess(code, processor)` manually without providing the third optional parameter ( omitting the filename )

https://github.com/wangjia184/svelte-parser/blob/master/src/worker.ts (not mine, not small)

**Expected behavior**
when inputing `const x : number = 1` the output to be `const x = 1`, even when no filename was provided.

**Additional context**
When you provide a weird filename like `" "` then the preprocessor will work and remove the type annotations.

### LINKED ISSUE #553: chore(deps): update stylus to 0.59.0
Level 4 CSS selector do not work on Stylus 0.55.0, (this was added on 0.58.0), since Stylus didn't reach 1.0 yet, major versions are still in middle numbers and being ignored, as of now, there is already a year gap since the latest release.

Update Stylus to 0.59.0

There is no alternative solutions, it just has to be updated, maybe add support for minor versions? 

Note: Important, level 4 selector are used in part of my components, they don't work without them.

**Additional context**
Nothing to add.

dummdidumm:
- breaking: remove TS mixed imports support, require TS 5.0 or higher
- breaking: require Svelte 4+, Node 18+
- breaking: add exports map
- fix: adjust type import
- fix: remove pug types from dependencies
- fix: allow TS filename to be undefined, fixes #488
- chore: replace jest with vitest
- chore: relax eslint config
- chore: bump peer deps, fixes #553, closes#635 
- chore: update CI, closes #638
- chore: get rid of magic-string as it's no longer used, bump node types
- docs: migration info


package.json
```diff
@@ -4,12 +4,20 @@
   "license": "MIT",
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
+  "exports": {
+    ".": {
+      "types": "./dist/index.d.ts",
+      "default": "./dist/index.js"
+    },
+    "./package.json": "./package.json"
```
benmccann, line 12: I don't think we need to do this in our packages anymore. It was a result of how our bundler plugins worked, but they've since been fixed

dummdidumm: better safe than sorry - I'm not sure if rollup-plugin-svelte and svelte-loader can handle it yet

benmccann: I'm pretty sure we updated them. Doesn't matter much, but figured I'd let you know

PR #641, opened 2024-06-14
fix: deprecate default export in favor of named export

dummdidumm:
This deprecates the default export in favor of the new named export `sveltePreprocess`. It's done to ensure a better interop between CJS and ESM without resorting to hacks in the future. It also enables people using `"module": "NodeNext"` in their `tsconfig.json` to import without type errors. The sub exports were also adjusted so that the transpiled TS output doesn't include `__importDefault` wrappers, which makes Node's static analysis miss those named exports.

Related: https://github.com/sveltejs/svelte-preprocess/issues/591#issuecomment-2166807552

### Tests

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -2,8 +2,8 @@ import { sveltePreprocess } from './autoProcess';
 
 // default auto processor
 // crazy es6/cjs export mix for backward compatibility
-// eslint-disable-next-line no-multi-assign
 /** @deprecated Use the named export instead: `import { sveltePreprocess } from 'svelte-preprocess'` */
+// eslint-disable-next-line no-multi-assign
 export default exports = module.exports = sveltePreprocess;
 
 // also export auto preprocessor as named export to sidestep default export type issues with "module": "NodeNext" in tsconfig.
--- a/src/index.ts
+++ b/src/index.ts
@@ -2,18 +2,22 @@ import { sveltePreprocess } from './autoProcess';
 
 // default auto processor
 // crazy es6/cjs export mix for backward compatibility
-
 // eslint-disable-next-line no-multi-assign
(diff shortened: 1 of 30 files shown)


src/index.ts
```diff
@@ -2,18 +2,22 @@ import { sveltePreprocess } from './autoProcess';
 
 // default auto processor
 // crazy es6/cjs export mix for backward compatibility
-
+/** @deprecated Use the named export instead: `import { sveltePreprocess } from 'svelte-preprocess'` */
 // eslint-disable-next-line no-multi-assign
 export default exports = module.exports = sveltePreprocess;
```
kaisermann, line 7: Do we still need this bizarre line? I don't quite remember why I had to do this whole exports/cjs/esm shenanigan. I guess it would be too early for another breaking change 😅

src/processors/babel.ts
```diff
@@ -26,3 +26,8 @@ export default (options?: Options.Babel): PreprocessorGroup => ({
     };
   },
 });
+
+// both for backwards compat with old svelte-preprocess versions
+// (was only default export once, now is named export because of transpilation causing node not to detect the named exports of 'svelte-preprocess' otherwise)
+export default babel;
```
kaisermann, line 32: Do we still need the default export here? IIRC these were re-exported with names in the package entrypoint.

src/index.ts
dummdidumm: yeah we need it because the default export is still present, and for interop reasons. I think it's best to not touch this line, just remove it in the next major.

src/processors/babel.ts
kaisermann: I find that a tad unlikely, but that's an unfortunately good point 🥲

kaisermann (approved):

PR #645, opened 2024-07-04
chore: remove all dependencies

benmccann:
these are such tiny utils it doesn't feel worth adding three dependencies (one has a transitive dependency of its own)


dummdidumm (approved):
Thank you!

PR #675, opened 2026-03-31
feat: support TypeScript 6

teemingc:
This PR adds TS6 support by ensuring source files adhere to TS6 rules and updating one of the tests to expect the appropriate contents for a `module: NodeNext` output

- [ ] This message body should clearly illustrate what problems it solves. If there are related issues, remember to reference them.

### Tests

- [ ] Run the tests with `npm test` or `pnpm test`


Keavon:
Thank you for this patch, I look forward to it landing and releasing to npm so it resolves the peer dependency resolution issue blocking our project's upgrade to TS 6. ❤️

src/transformers/typescript.ts
```diff
@@ -64,16 +64,17 @@ function getCompilerOptions({
   const compilerOptions: CompilerOptions = {
     target: ts.ScriptTarget.ES2015,
     ...convertedCompilerOptions,
-    // force module(resolution) to esnext and a compatible moduleResolution. Reason:
-    // transpileModule treats NodeNext as CommonJS because it doesn't read the package.json.
```
teemingc, line 68: I don't think this is the case now but maybe @dummdidumm can verify

dummdidumm: What do you mean by "now"? In TS6? Maybe. But very likely in 5.x - did you check?

For best backwards compat we should check the TS version and use the old logic for TS < 6

teemingc: Yeah, you're right. It's not an issue for TS6 but I checked TS 5.x and we do need the backwards compat. Just added it

src/autoProcess.ts
```diff
@@ -58,7 +58,7 @@ export function sveltePreprocess(
   {
     aliases,
     markupTagName = 'template',
-    sourceMap = process?.env?.NODE_ENV === 'development' ?? false,
+    sourceMap = process?.env?.NODE_ENV === 'development' || false,
```
teemingc, line 61: Fixes this TypeScript error
> Right operand of ?? is unreachable because the left operand is never nullish.

test/transformers/typescript.test.ts
```diff
@@ -139,7 +139,9 @@ describe('transformer - typescript', () => {
 
       const { code } = await preprocess(tpl, opts);
 
-      expect(code).toBe(`<script lang="ts" context="module"></script>`);
+      expect(code).toBe(
+        `<script lang="ts" context="module">export {};\n</script>`,
```
teemingc, line 143: Empty modules created with `NodeNext` always have `export {}` while `ESNext` simply has `"use strict"`

dummdidumm (approved):
thank you!

